PR #9, opened 2016-05-12
obvious-ci -> linux-anvil

jakirkham:
Fixes https://github.com/conda-forge/docker-images/issues/8

Renames the base image from `obvious-ci` to `linux-anvil` based on this [discussion]( https://github.com/conda-forge/docker-images/issues/8#issuecomment-218662290 ). Will need to fix up Docker Hub after this change.


jakirkham:
Should we delete `condaforge/obvious-ci`? It wasn't actually used for anything.

jjhelmus:
I don't think it hurts keeping it there but since it is not and was neved used I have no problems removing it.

jakirkham:
Went ahead and removed it.

PR #13, opened 2016-05-20
Configure CircleCI

jakirkham:
Configure the repo so that it can perform CI with CircleCI. This will let us test the builds of our images quickly and will avoid having to build them by hand locally. Hopefully with this change we can have a better idea of whether a proposed change is working without needing to spend cycles building it locally. To start of the CI, this just tries to get the image built. In the future, this could be used to test image(s) built to make sure they behave ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,4 @@
+[![CircleCI](https://circleci.com/gh/conda-forge/docker-images/tree/master.svg?style=svg)](https://circleci.com/gh/conda-forge/docker-images/tree/master)
+
 # docker-images
 Repository to host the Docker images files used in conda-forge
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,11 @@
+machine:
+  services:
+    - docker
+
+dependencies:
+  override:
+    - docker info
+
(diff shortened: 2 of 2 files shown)


jakirkham:
As FYI, one might notice a line after each layer that seems worrisome. Though this isn't actually causing us any problems.

```
Error removing intermediate container 1112e5148a62: no such id: 1112e5148a6286e6ee4ec8a9770cf4fc791f3adff77d85fe59709255e793359d
```

This is basically some weird behavior due to CircleCI's use of run containers without privileges (completely reasonable by them). Though they seem to be forcing `--rm`, which is causing this issue. There is discussion on CircleCI's forum about this [issue]( https://discuss.circleci.com/t/docker-error-removing-intermediate-container/70/14 ). We should keep an eye on it, but personally I'm not worried about it.

jakirkham:
Going to run with this as we need some kind of CI here. We can always swap it out for a different kind later. If anyone has strong objections with using CircleCI for this, please let me know and I would be happy to find an appropriate resolution.

Issue #14, opened 2016-05-20
DockerHub autobuilds failing


jakirkham:
The DockerHub builds are having some sort of checksum error when installing some things. As this is not happening with CircleCI and I am unable to reproduce it locally, it seems like a DockerHub issue. Added the following upstream issue ( https://github.com/docker/hub-feedback/issues/675 ). Let's hold on new changes for a bit.

If DockerHub remains unstable in these sorts of ways (sadly not the first issue I have encountered with automated builds on DockerHub), we could try [quay.io]( https://quay.io/ ) or we could just push images from CircleCI. If we try the former, we can mirror images on DockerHub to avoid having to switch a bunch of links.

jakirkham:
Alright, DockerHub is still choking on this image. So, I'm setting up [quay.io]( https://quay.io/ ). Have only heard good things about it. Figured we can give it a go. If needed, we can mirror the image back to DockerHub. Invited everyone that was an owner before to join and created an image for `linux-anvil`.

jakirkham:
Adding a PR ( https://github.com/conda-forge/docker-images/pull/16 ), which will give us a `Dockerfile` that we can use to mirror the [quay.io]( https://quay.io/ ) builds that are working quite nicely. We will also need to set up a build trigger to get this to work automatically. I have already activated that on the DockerHub end, but will investigate the changes required on the quay.io end.

jakirkham:
Setup the build trigger using a webhook and have tested that it works. If we are ok with this plan, we are good to go.

jakirkham:
I guess "resolves" now triggers closes. In any event, this does seem to be resolved.

PR #16, opened 2016-05-27
Add a Dockerfile to mirror quay.io

jakirkham:
Partially resolves https://github.com/conda-forge/docker-images/issues/14

This adds a `Dockerfile` that merely pulls the [image]( https://quay.io/repository/condaforge/linux-anvil ) we are building on [quay.io]( https://quay.io/ ). The purpose of doing this is we could change our Docker Hub build to use this instead of building the image from scratch. Thus we will not need to go through and change our `FROM` commands, but we can actually build our image again. This is increasingly important as we are needing to make changes to it.

Additionally, I have enabled a webhook on the Docker Hub side and configured quay.io to post using that webhook after pushing an image. I just tested this system and it appears to work. So, when we merge something here it will be rebuilt on quay.io and once built will trigger a build on DockerHub that will pull in this image.

The last piece that we need to do after merging this is to change the directory that DockerHub uses for the `Dockerfile` to point to this directory that we are adding.

cc @pelson @jjhelmus @msarahan @ocefpaf

### DIFF
--- a/linux-anvil-mirror/Dockerfile
+++ b/linux-anvil-mirror/Dockerfile
@@ -1,0 +1,1 @@
+FROM quay.io/condaforge/linux-anvil:latest


jakirkham:
Updated the description above with more recent things I have tried. Basically, this is working and we are ready to go.

jakirkham:
Alright, as there is at least one vote in favor and I *need* to make some changes to the image for end users, I'm going to merge this. I am still willing to discuss this further, but the most immediate problem is fixing the build failures somehow.

ocefpaf:
Two now :wink:

jakirkham:
Thanks @ocefpaf.

Sorry to drive ahead here. I do want to talk about this more, but also want things to be working here and know you guys are otherwise engaged for the most part. We can make it a topic for the next meeting if you would like.

jakirkham:
Were you able to get access to quay.io ok?

jakirkham:
FYI, this works! 😄

screen shot 2016-05-27 at 10 12 59

jakirkham:
Should add that if in the future we want a different build service (though I'm pretty happy with quay now), this nicely decouples the build from the Docker Hub hosted image so that we can easily change it later.

PR #17, opened 2016-05-27
Bust the cache in linux-anvil

jakirkham:
This adds a step that always busts the cache. Based on a [workaround]( https://github.com/docker/docker/pull/10682#issuecomment-178810479 ) offered in-place of the unaccepted `NOCACHE` command. The purpose of this step is to ensure that we are not re-using cached downloads. We accomplish this by `ADD`ing a file that we download from a URL that always changes. Namely we add a file that contains the current time. As a result, we will have a different hash for this layer each time and will need to generate every subsequent layer from scratch.

### DIFF
--- a/linux-anvil/Dockerfile
+++ b/linux-anvil/Dockerfile
@@ -6,6 +6,9 @@ MAINTAINER conda-forge <conda-forge@googlegroups.com>
 # Set an encoding to make things work smoothly.
 ENV LANG en_US.UTF-8
 
+# Add a timestamp for the build. Also, bust the cache.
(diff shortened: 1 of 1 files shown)


jakirkham:
As quay.io does do proper caching of layers (unlike Docker Hub) and they don't provide a setting to disable this, we require a step that will bust the cache before downloading anything. That is exactly what this step does. More detailed description in the OP.

jakirkham:
Going to put this in so we can have a newer `git` for people. ( https://github.com/conda-forge/docker-images/pull/12 ) Also, this is a pretty trivial change so am hoping there isn't any real controversy here. Again happy to discuss with people as their schedule allows.

PR #27, opened 2016-08-17
Update to `tini` version `0.10.0`

jakirkham:
Update to `tini` version `0.10.0` so that the license is included. This ensures we are compliant with the terms of the MIT license that `tini` is under.

**Changelog:**
```
- New `-l` option to show license and exit.
```

cc @tylere


jakirkham:
We should really be including the license for this and that is the only change made in this version of `tini`. Going to go ahead and merge this so we are compliant. Hope that is ok.

Issue #6, opened 2016-05-01
Make sure that `USER` is defined


jakirkham:
We need to tweak the image to make sure `USER` is defined. On a normal Linux system this would be defined, but that is not the case here. However, this is very easy to change. This is coming up because of this proposed recipe ( https://github.com/conda-forge/staged-recipes/pull/477 ).

jakirkham:
Fixed in [current `master`]( https://github.com/conda-forge/docker-images/commit/1b09235d44a443adfe4a4f1d9081d38795558861 ). Presumably fixed by PR ( https://github.com/conda-forge/docker-images/pull/28 ).

PR #28, opened 2016-08-17
Fix prompt

jakirkham:
Previously the `PS1` prompt was set correctly by `bash` when the container started.

However, now `conda` mucks with `PS1` before `bash` starts in activation. So `bash` won't change the prompt afterwards. Further `conda` deactivation will hang.

This sets the prompt before `conda` activates so that there is a prompt and `conda` deactivation works. This is primarily of value for offline use.

Tested for local use interactively. Also tested building `pycrypto` with the Docker build script using it. Both worked fine.

### DIFF
--- a/linux-anvil/entrypoint
+++ b/linux-anvil/entrypoint
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -il
 
 # Source everything that needs to be.
 . /opt/docker/bin/entrypoint_source


jakirkham:
As this is a simple change that I have already tested some, I'd like to see if we can get it out there. Am going to test on a feedstock to make sure it works.

jakirkham:
Hope this is all ok.

jakirkham:
Tested and verified to still work in [build]( https://circleci.com/gh/conda-forge/line_profiler-feedstock/19 ).

jakirkham:
As this made the shell a login shell, it probably fixed issue ( https://github.com/conda-forge/docker-images/issues/6 ). However, I did not check at the time. Suffice it to say that issue is resolved.

PR #29, opened 2016-08-22
Validate Miniconda download with checksum

jakirkham:
This picks a particular version of Miniconda to install (still the most recent one available) and installs it. Also, it verifies that its checksum matches what we expect it too. Not that we have any reason to think these wouldn't match. Still it is good practice to be using checksums to verify things that we have downloaded.

We need to pick a particular Miniconda version so that we can be sure about the checksum we are getting. If we were downloading latest, the checksum would change when there is a new Miniconda release. This shouldn't be a problem when we download a specific version.

In the future, we will need to explicitly update our Miniconda installs to new versions as needed. However, in practice, we may not need to do this very often as we upgrade everything in the Docker build anyways. Plus, we are switching over to conda-forge packages for nearly everything that would come included.

### DIFF
--- a/linux-anvil/Dockerfile
+++ b/linux-anvil/Dockerfile
@@ -48,8 +48,10 @@ RUN curl -s -L -O https://github.com/krallin/tini/releases/download/v0.10.0/tini
     mv tini /usr/local/bin
 
 # Install the latest Miniconda with Python 3 and update everything.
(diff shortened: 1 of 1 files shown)


jakirkham:
As this is a harmless change that improves the quality of our image and it has been sitting for awhile without much response, went ahead with it. Hope that is ok. Feel free to share thoughts if you have any.

PR #32, opened 2016-08-22
Drop system `gcc` (have devtoolset already)

jakirkham:
This proposes we drop the system `gcc`. As we already include the devtoolset compiler, there is really no need for this other compiler to also be installed. After all the system `gcc` is not even used. Dropping the system `gcc` would result in a ~5% size reduction in the final image. Have tested some simple C and C++ programs to verify this doesn't have any effect.


jakirkham:
As this is a harmless change that improves the quality of our image and it has been sitting for awhile without much response, went ahead with it. Hope that is ok. Feel free to share thoughts if you have any.

PR #41, opened 2016-10-05
Install Xvfb

jschueller:
Could be needed to run gui tests:
https://github.com/conda-forge/staged-recipes/pull/1684

### DIFF
--- a/linux-anvil/Dockerfile
+++ b/linux-anvil/Dockerfile
@@ -21,6 +21,7 @@ RUN yum update -y && \
                    sudo \
                    tar \
                    which \
+                   xorg-x11-server-Xvfb \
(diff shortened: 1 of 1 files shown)


jakirkham:
Why not just use `yum_requirements.txt`?

jakirkham:
Please see this [recipe]( https://github.com/conda-forge/qtconsole-feedstock/tree/421859b54e7ead2c2193971aefda86c8e55dbc57/recipe ) as an example.

jschueller:
I did not know that.

jakirkham:
No worries. FWIW I'm not saying this shouldn't be here. Just saying that if there is something not here that would be the thing to try first.

PR #42, opened 2016-10-31
Fix Miniconda download

jakirkham:
Fixes the download URL for Miniconda as has been done for Travis and AppVeyor.

xref: https://github.com/conda-forge/conda-smithy/pull/347
xref: https://github.com/conda-forge/conda-smithy/pull/348


jakirkham:
The Docker image is not affected like the other CIs as we use a fixed version of the Miniconda installer. Still as this does work and is the right way to do it. Going to go ahead and switch to `https`. Shouldn't have any other effect on the end result.

PR #44, opened 2017-01-19
Switch to the USNO's time service

jakirkham:
It appears that timeapi.org has been taken offline and the domain is up for grabs. To still ensure the cache is busted at the beginning of the Docker build, this switches to the USNO's time service to get the time since the last epoch. This should allow us to still bust the cache and save the time when the image was built.


jakirkham:
Going to get this out of the way to fix the failing builds.

PR #47, opened 2017-04-18
Install conda build and deployment tools directly

jakirkham:
Drops the last vestige of Obvious-CI from the docker-image and handles this install step directly. Thus removing the need for this Obvious-CI script.

### DIFF
--- a/linux-anvil/Dockerfile
+++ b/linux-anvil/Dockerfile
@@ -73,10 +73,9 @@ RUN curl -s -L https://repo.continuum.io/miniconda/Miniconda3-4.2.12-Linux-x86_6
     conda update --all --yes && \
     conda clean -tipy
 
-# Install Obvious-CI.
+# Install conda build and deployment tools.
 RUN export PATH="/opt/conda/bin:${PATH}" && \
-    conda install --yes obvious-ci && \
-    obvci_install_conda_build_tools.py && \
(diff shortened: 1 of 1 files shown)


linux-anvil/Dockerfile
```diff
@@ -73,10 +73,9 @@ RUN curl -s -L https://repo.continuum.io/miniconda/Miniconda3-4.2.12-Linux-x86_6
     conda update --all --yes && \
     conda clean -tipy
 
-# Install Obvious-CI.
+# Install conda build and deployment tools.
 RUN export PATH="/opt/conda/bin:${PATH}" && \
-    conda install --yes obvious-ci && \
-    obvci_install_conda_build_tools.py && \
+    conda install --yes --quiet conda-build anaconda-client jinja2 setuptools && \
```
jakirkham: Given this script is a thin wrapper around an [installation]( https://github.com/pelson/Obvious-CI/blob/v0.6.1/scripts/obvci_install_conda_build_tools.py#L17-L18 ) of these [packages]( https://github.com/pelson/Obvious-CI/blob/v0.6.1/scripts/obvci_install_conda_build_tools.py#L12 ), we are better served by doing that ourselves. That way, we can change it directly if we need to in the future.

jakirkham:
Could you please take a look @conda-forge/core?

jjhelmus:
LGTM

PR #48, opened 2017-06-27
Install and use conda-forge's tini

jakirkham:
Instead of downloading the pre-built binary for `tini`. Install our own `conda-forge` package for `tini`, which we build from source.


scopatz:
Thanks!

PR #53, opened 2018-01-06
Ensure that the `conda` user owns home

jakirkham:
In case the `conda` user's home directory already exists (as happens when something gets mounted to that location), make sure that the home directory still ends up being owned by the `conda` user in the end and all of its contents.


jakirkham:
NVM. This appears to be happening already as long as `HOST_USER_ID` is set correctly.

jakirkham:
Turns out this is still an issue in some cases. PR ( https://github.com/conda-forge/docker-images/pull/53 ) will fix it though.

PR #55, opened 2018-01-06
Switch to the `conda` user's home directory

jakirkham:
Touch up to PR ( https://github.com/conda-forge/docker-images/pull/36 ).

Instead of dropping into `/`, it makes more sense to drop into `/home/conda`. As we use absolute paths throughout, this is largely a cosmetic touch. However it would help if the current working directory is a place that allows us to create things without issues.


jakirkham:
Turns out there are some [non-cosmetic, beneficial effects]( https://github.com/conda-forge/gawk-feedstock/issues/1#issuecomment-355782567 ) as well.

bollwyvl:
Cool!

PR #68, opened 2018-06-16
move chown to dockerfile

### LINKED ISSUE #63: chmod in entrypoint takes ages

sdvillal:
closes #63

Do we have a good way to add a test touching a file in /opt/conda?


jakirkham:
Could you please take a look, @gqmelo?

PR #69, opened 2018-07-05
Use new time server to bust cache

jakirkham:
Appears that USNO's time server redirects to `https`, but their certificates are invalid for some reason. Probably a consequence of their recent move to `https`. Would expect them to fix this in the future. For now, switch to a new time server to bust the cache. This provides the output in JSON with a few formats as opposed to a simple string, but is otherwise just text. Should work for our needs.


scopatz:
#67 seems to remove the timestamp, not fix it. Merging this small change for the moment.

jakirkham:
Thanks.

FWIW was not opposed to getting that PR merged, but AFAICT I was the only other core person beside @isuruf to agree with it. So was waiting for more buy-in. Plus going to AFK for a few days. So was taking the cautious approach generally.

PR #72, opened 2018-08-22
Use force when removing `miniconda.sh`

jakirkham:
As this is a file downloaded by `root`, the system will ask us to confirm our decision to remove the file. Since we are running in batch, it simply assumes the answer is no and proceeds. As a result, a copy of `miniconda.sh` (weighing in at about 34MB) is getting packaged in our Docker image increasing the image size.

To fix this, start using `rm -f` to forcibly remove `miniconda.sh`. This way when Docker goes to commit this layer, `miniconda.sh` won't be present and so will not be included. Should help improve image size and cutdown on clutter.

### DIFF
--- a/scripts/run_commands
+++ b/scripts/run_commands
@@ -20,7 +20,7 @@ echo 'conda ALL=NOPASSWD: /usr/bin/yum' >> /etc/sudoers
 curl -s -L https://repo.continuum.io/miniconda/$condapkg > miniconda.sh
 openssl md5 miniconda.sh | grep $conda_chksum
 bash miniconda.sh -b -p /opt/conda
-rm miniconda.sh
+rm -f miniconda.sh
 touch /opt/conda/conda-meta/pinned
 ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh
 source /opt/conda/etc/profile.d/conda.sh


mbargull:
The change is fine. But I think this just happened because the shebang is incorrect. It should be just `#!/bin/bash -l` and not interactive, right?!

jakirkham:
I don't think they are related. The file was removed because of the [sticky bit]( https://www.linuxnix.com/sticky-bit-set-linux/ ).

The `-i` is intentional otherwise the prompt and some other things (it's been a few years since I looked) get messed up. ( https://github.com/conda-forge/docker-images/pull/28 )

mbargull:
Pretty sure they are related :wink:. I don't really know the reason of it, but for some distributions in the default configuration `rm` will ask for confirmation if it's an interactive shell. CentOS is one of those ;).
And we needn't care about the prompt here since it's only the script that gets run during the build and not an entry point.

jakirkham:
Mainly I've seen the prompt when `root` creates a file, which gets the sticky bit. Normal users don't see this. For example...

<details>

```
$ docker run --rm -it condaforge/linux-anvil
[conda@e84699c533c9 ~]$ mkdir test_dir && cd test_dir
[conda@e84699c533c9 test_dir]$ ls
[conda@e84699c533c9 test_dir]$ touch myfile.txt
[conda@e84699c533c9 test_dir]$ ls
myfile.txt
[conda@e84699c533c9 test_dir]$ rm myfile.txt
[conda@e84699c533c9 test_dir]$ ls
```

</details>

<br>

Didn't realize this before, but apparently `root` has `alias rm='rm -i'` in `.bashrc`. From the man `rm` man page.

<details>

```
-i          Request confirmation before attempting to remove each file, regardless of the file's
             permissions, or whether or not the standard input device is a terminal.
```

</details>

<br>

Makes me wonder why it ever worked without `-f`.

You're probably right that we don't need to worry about the prompt in batch install script. It is useful in the entrypoint. Given `-il` seems to be the right way to get `conda` working normally, there shouldn't be any harm in sticking with it even for batch install scripts.

On a side note, most of these fixes that went in recently are needed to clean up after moving contents from the `Dockerfile` to an install script to support PPC in addition to x86-64. So it's not surprising a few fixes were needed.

PR #73, opened 2018-10-13
Download and cache new compiler packages

jakirkham:
Make sure the new compiler packages are downloaded and cached to make it easier to install quickly at build time.

cc @mariusvniekerk


mariusvniekerk:
Do we want to update the yum things we install to not install sudo etc?

jakirkham:
Not sure I follow. We need `sudo` to install stuff as a non-`root` user in the feedstocks. Particularly when `yum_requirements.txt` is used.

jakirkham:
Any other thoughts?

jakirkham:
Thoughts @conda-forge/core?

mariusvniekerk:
Lgtm

PR #77, opened 2018-12-30
Drop X11 reqs from new compiler Docker image

jakirkham:
All of these are either conda-forge packages that we should be using and/or are using or there are CDTs we should be using for these. Given this, go ahead and drop these X11 packages from the new compiler Docker image. Admittedly this is a breaking change for some packages, but one that we are already adapting in many places.

### DIFF
--- a/linux-anvil-comp7/Dockerfile
+++ b/linux-anvil-comp7/Dockerfile
@@ -19,12 +19,7 @@ RUN yum update -y && \
                    patch \
                    sudo \
                    tar \
-                   which \
-                   libXext-devel \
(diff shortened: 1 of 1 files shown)


jakirkham:
Thoughts @conda-forge/core?

jakirkham:
Actually this is even safer than the description in the OP suggests. It seems the new compilers, which are used exclusively in the `Dockerfile` modified in this PR, don't really look at system libraries at all. In fact they will complain about X11 packages missing even when installed by `yum`.

scopatz:
Thabjs @jakirkham!

PR #81, opened 2019-02-14
Update entrypoint

hiaselhans:
Use the non-interactive version of cp to avoid asking to replace ".condarc"


scopatz:
Thanks @hiaselhans

PR #99, opened 2019-06-27
Use `docker build --squash...`

jakirkham:
Fixes https://github.com/conda-forge/docker-images/issues/82

This enables [Docker's experimental features]( https://github.com/docker/docker-ce/blob/554e31a7e6544d17ef74d5656c0e6d1af590fbd0/components/cli/experimental/README.md#use-docker-experimental ) on Travis CI, which are needed to run `docker build` with [the `--squash` flag]( https://github.com/docker/docker-ce/blob/554e31a7e6544d17ef74d5656c0e6d1af590fbd0/components/cli/docs/reference/commandline/build.md#squash-an-images-layers---squash-experimental ). When using this flag with `docker build`, all layers are squashed together into one layer at the end. This history itself is retained, but none of the previous layers contain anything addition. Though the base image (used in `FROM`) is retained in the history with its original size.

To get a sense of what this would provide, I tried building the `condaforge/linux-anvil-comp7` image. Currently the image size is 2.45GB. After building with `--squash`, the image is 1.74GB. Overall this is a 0.71GB reduction in size or roughly 30%. All just by flipping a switch and adding a flag!

This flag is great because the benefit is immediately extended to all of the images we maintain here. Also it carries forward to whatever changes we make to the images here. Plus we no longer need to feel a need to jam all commands into one `RUN` (unless we like that style for some reason). So all around this a nice win for both image size and maintainability here with very little effort. 😄

cc @epruesse


scopatz:
This is great @jakirkham!

epruesse:
+1 This is very nice!

mbargull:
I'm not aware of any issues around `--squash`, so I guess/hope this change is fine. Generally, I think multi-staged builds would ease things comparably and might lend themselves better for other future changes. However,

> benefit is immediately extended to all of the images we maintain here

is a nice thing to have for now!

----

> To get a sense of what this would provide, I tried building the `condaforge/linux-anvil-comp7` image. Currently the image size is 2.45GB. After building with `--squash`, the image is 1.74GB. Overall this is a 0.71GB reduction in size or roughly 30%. All just by flipping a switch and adding a flag!

Taking a look at https://github.com/conda-forge/docker-images/blob/3bc379c9cbc4c3ae11ed7e4683f9a76cf7d22418/linux-anvil-comp7/Dockerfile#L53-L54 , I'm pretty sure the culprit is in the overzealous `chgrp -R`/`chmod -R`, i.e., those commands change (and thus duplicate in terms of layer contents) already existent files. I'll see if we can do the `chgrp`/`chmod` more selectively.
(I know we don't have to fix that if `--squash` works perfectly as intended (or we use multi-stage builds). But since the change should be minimal, I guess we could fix it anyway, just in case.)

PR #103, opened 2019-06-28
Update conda-build install

jakirkham:
Ensures all the [other tools]( https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/building-conda-packages.html#other-tools ) recommended for use with conda-build are installed. Also drops explicit install of some things like `jinja2`, which are `conda-build` requirements. Some other general tidying.


scripts/run_commands
```diff
@@ -43,7 +43,8 @@ conda update --all --yes
 conda clean -tipy
 
 # Install conda build and deployment tools.
-conda install --yes --quiet conda-build anaconda-client jinja2 setuptools git patch python=3.7
+conda install --yes --quiet bzip2 git patch
+conda install --yes --quiet python=3.7 setuptools conda-build anaconda-client
```
scopatz: one install command is better than two, so we don't solve twice

mbargull: Yes, I agree, a single command is preferable. To retain the indent of https://github.com/conda-forge/docker-images/pull/103/commits/72431c84e8027be9a0e9c8a6ab76e22eadf9b9ba on readability/maintainability, we can just use a multi-line command:
```sh
conda install --yes --quiet \
   bzip2 git patch \
   python=3.7 setuptools \
   conda-build anaconda-client
```
The comment above the command may be updated to reflect that structure, too.

----
Regarding `bzip2`, I believe https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/building-conda-packages.html#other-tools is just not accurate on this. I'm not aware that `conda-build` uses `bzip2` directly, rather than via `python` (or `libarchive`).
My best guess is that either some ancient `conda-build` version did indeed use `bzip2` directly or someone mixed up `conda-build`'s and `Miniconda`'s requirements (the latter (currently) does require `bzip2`, IIRC). IMHO, we shouldn't add it here. (@msarahan would know better, of course.)

----
+1 on the `jinja2` removal!

scopatz:
This has some conflicts now

scripts/run_commands
jakirkham: Have combined the commands and split over lines as suggested.

jakirkham:
Think I've addressed the aforementioned concerns. Please take another look. 🙂

mbargull:
LGTM, apart from the `bzip2` addition.
I'm confident that https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/building-conda-packages.html#other-tools is wrong on that and we don't need the `bzip2` package.

jakirkham:
If that's wrong, then the docs should be updated. Would you like to take that on?

Have no objection to the line changing here once the docs are updated.

Given this, it sounds like we can merge and this can be updated as the docs update in the future.

mbargull:
I was just going to open a PR for the docs (change at https://github.com/mbargull/conda-build/commit/be0e5bb32cecbda3055ad38233762972e5e67244). Then I read the text: (emphasis mine)
> Some environments initially lack tools such as bzip2, patch, or Git
> that **may be needed for some build workflows**.

I doesn't even list those tools as "recommended" or needed for `conda-build`. Meaning, you could also, e.g., write "tools such as xz, patch, or Git \[...\] for some build workflows". It doesn't really matter.

mbargull:
Searching the docs for `bzip2` only mentions it in one other place in a different context (as a library):
https://github.com/conda/conda-build/search?l=reStructuredText&q=bzip2
https://github.com/conda/conda-build/blob/93eeae6929224a2b61fd35119de06f22376a6e94/docs/source/resources/define-metadata.rst#L739

So okay, for consistency's sake, `bzip2` should be removed from https://github.com/conda/conda-build/blob/ca4dab286e20bfa7af9810e7f9ea7a7d2c790dec/docs/source/user-guide/tutorials/building-conda-packages.rst.
But that is totally irrelevant for this PR, so please just remove it here.

mbargull:
xref: https://github.com/conda/conda-build/pull/3625

jakirkham:
Thanks for doing that. Have updated this as well.

PR #107, opened 2019-07-01
 Make subsequent chgrp/chmod more selective (reduce image size)

mbargull:
Addresses gh-82

Details:
@mbargull in https://github.com/conda-forge/docker-images/pull/99#issuecomment-507417269 :
> Taking a look at https://github.com/conda-forge/docker-images/blob/3bc379c9cbc4c3ae11ed7e4683f9a76cf7d22418/linux-anvil-comp7/Dockerfile#L53-L54 , I'm pretty sure the culprit is in the overzealous `chgrp -R`/`chmod -R`, i.e., those commands change (and thus duplicate in terms of layer contents) already existent files. I'll see if we can do the `chgrp`/`chmod` more selectively.
(I know we don't have to fix that if `--squash` works perfectly as intended (or we use multi-stage builds). But since the change should be minimal, I guess we could fix it anyway, just in case.)

### DIFF
--- a/linux-anvil-aarch64/Dockerfile
+++ b/linux-anvil-aarch64/Dockerfile
@@ -54,7 +54,7 @@ RUN source /opt/conda/etc/profile.d/conda.sh && \
         $_channel::libgfortran-ng \
         $_channel::libstdcxx-ng && \
     conda remove --yes --quiet -n test --all && \
-    conda clean -tsy && \
+    conda clean -tiy && \
     find /opt/conda ! -group lucky -exec chgrp -h lucky {} + -exec chmod u=g {} +
 
 # Add a file for users to source to activate the `conda`
--- a/linux-anvil-aarch64/Dockerfile
+++ b/linux-anvil-aarch64/Dockerfile
@@ -26,6 +26,7 @@ RUN localedef -i en_US -f UTF-8 en_US.UTF-8
 RUN yum update -y && \
     yum install -y \
         bzip2 \
+        findutils \
         sudo \
         tar \
         which && \
@@ -54,8 +55,7 @@ RUN source /opt/conda/etc/profile.d/conda.sh && \
         $_channel::libstdcxx-ng && \
     conda remove --yes --quiet -n test --all && \
     conda clean -tsy && \
-    chgrp -R lucky /opt/conda && \
-    chmod -R g=u /opt/conda
+    find /opt/conda ! -group lucky -exec chgrp -h lucky {} + -exec chmod u=g {} +
 
 # Add a file for users to source to activate the `conda`
 # environment `base`. Also add a file that wraps that for
--- a/linux-anvil-cuda/Dockerfile
+++ b/linux-anvil-cuda/Dockerfile
@@ -62,7 +62,7 @@ RUN source /opt/conda/etc/profile.d/conda.sh && \
         defaults::libgfortran-ng \
         defaults::libstdcxx-ng && \
     conda remove --yes --quiet -n test --all && \
-    conda clean -tsy && \
+    conda clean -tiy && \
     find /opt/conda ! -group lucky -exec chgrp -h lucky {} + -exec chmod u=g {} +
 
 # Download and cache CUDA related packages.
--- a/linux-anvil-cuda/Dockerfile
+++ b/linux-anvil-cuda/Dockerfile
@@ -35,6 +35,7 @@ RUN rpm -e --nodeps --verbose gcc gcc-c++
 RUN yum update -y && \
     yum install -y \
         bzip2 \
+        findutils \
         sudo \
         tar \
         which && \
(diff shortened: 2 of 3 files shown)


linux-anvil-aarch64/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN localedef -i en_US -f UTF-8 en_US.UTF-8
 RUN yum update -y && \
     yum install -y \
         bzip2 \
+        findutils \
```
mbargull: `findutils` is already installed, so this is a no-op. However, I like to be explicit about what we use here.

linux-anvil-aarch64/Dockerfile
```diff
@@ -53,9 +54,8 @@ RUN source /opt/conda/etc/profile.d/conda.sh && \
         $_channel::libgfortran-ng \
         $_channel::libstdcxx-ng && \
     conda remove --yes --quiet -n test --all && \
-    conda clean -tsy && \
-    chgrp -R lucky /opt/conda && \
-    chmod -R g=u /opt/conda
+    conda clean -tiy && \
+    find /opt/conda ! -group lucky -exec chgrp -h lucky {} + -exec chmod u=g {} +
```
mbargull: This runs `chgrp`/`chmod` only for those files that haven't already been `chgrp`'d for the `lucky` group => we avoid touching those files created by `scripts/run_commands` in the lower layer.

jakirkham:
Is this needed if we already squash the layers? ( https://github.com/conda-forge/docker-images/pull/99 )

mbargull:
> (I know we don't have to fix that if `--squash` works perfectly as intended (or we use multi-stage builds). But since the change should be minimal, I guess we could fix it anyway, just in case.)

It's a minimal change and it's nice to have it in place in case we encounter problems with squashing, IMHO.

linux-anvil-cuda/Dockerfile
```diff
@@ -35,6 +35,7 @@ RUN rpm -e --nodeps --verbose gcc gcc-c++
 RUN yum update -y && \
     yum install -y \
         bzip2 \
+        findutils \
```
scopatz: Please use the conda-forge package: https://anaconda.org/conda-forge/findutils

jakirkham:
It's more a concern about readability. This command requires a bit more thought to understand. Unless it is delivering measurable value, I'd prefer to leave it out. Sorry.

linux-anvil-cuda/Dockerfile
mbargull: I'd rather not, for now.
- `conda install findutils` would only add unneeded baggage.
- `yum install findutils` is a no-op current since it is already installed.
- It is in the same league as `bzip2`, `tar` => we only install those to run the Miniconda install, but can (or rather should, IMO) ditch them if/when we use multi-stage builds (where we needn't run the installer at all).

So, if you feel strongly about not running `yum install findutils`, then I'd rather say we neither run `yum install findutils` nor `conda install findutils`: The images will still build fine since `find` is already part of the base image, we would just be a little less verbose/explicit ( https://github.com/conda-forge/docker-images/pull/107/files#r299215307 ).

mbargull:
> This command requires a bit more thought to understand.

Knowing how `find` works, it's *comparably* easy to understand. Though I get that not everyone is familiar with it. But I'd rather add an explanatory comment than not have it, TBH.

jakirkham:
> Unless it is delivering measurable value...

Is it reducing the size?

mbargull:
> Is it reducing the size?

With `--squash`, no. Without `--squash`, yes, since it addresses the cause of gh-82, hence avoids adding the redundant ~ 600 MiB for that layer. My reasoning is just:
> It's a minimal change and it's nice to have it in place in case we encounter problems with squashing, IMHO.

----
That being said, I tested this locally today and, surprisingly, have not been able to reproduce gh-82 with my up-to-date Docker version. Turns out you can reproduce it with `17.09.0` (and verify this PR works as intended) but something was changed in `17.10.0` such that all newer versions don't store those redundancies. Travis' default Docker version is of course `17.09.0` XD.

Overall, I'd like to update the image build files to use multi-stage builds at some point since it can simplify and cleanup everything considerably. But I'll hold up on that just a little bit since there is at least [one change](https://github.com/docker/docker-ce/commit/9e3a6542abc854019bcb0c4a86fdff591fdbd6df#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR68) that will be part of a [very soon to be released](https://github.com/docker/docker-ce/pull/612/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR6) Docker version.

jakirkham:
Sorry, I'm going to stick to my original opinion. I'd prefer to not merge this PR as-is.

We do not need this with squashing. Also based on your experiment, it sounds like Docker may be doing this for us (even without squashing). Given this, adding more complex code for problems we do not have does not sound like a good idea. So there does not appear to be a compelling reason for these changes.

mbargull:
> more complex code

To me, both variants look comparably cryptic without knowing the context ¯\\_\(ツ)\_/¯
```diff
-    chgrp -R lucky /opt/conda && \
-    chmod -R g=u /opt/conda
+    find /opt/conda ! -group lucky \
+        -exec chgrp -h lucky {} + \
+        -exec chmod u=g {} +
```
> Sorry, I'm going to stick to my original opinion. 

Likewise. Can't always agree ;), no hard feelings. Happy to let a third party decide.

linux-anvil-aarch64/Dockerfile
scopatz: Sure, but I want to keep this list as small as possible

PR #97, opened 2019-06-26
Ensure locales work

jakirkham:
Applies the same changes we have in the ARM and POWER images to our other images to ensure the locales are working correctly.

Note: This also comes up in [this SO answer]( https://unix.stackexchange.com/a/140517 ).

Checklist
* [ ] Used a fork of the feedstock to propose changes
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #108, opened 2019-07-03
 Test for working en_US.UTF-8 locale 

mbargull:
Do we really need the changes from gh-83 and gh-97?
It seems like https://github.com/CentOS/sig-cloud-instance-images/issues/71 is a resolved issue and the locale is working, isn't it?
I'd rather like us to test whether the locale is working and only apply fixes if it really is not.
I couldn't find a reference to a recent issue in gh-83. @mariusvniekerk, @hmaarrfk, can you point me to failing cases such that we can incorporate them in the tests?

Also, @mariusvniekerk, gh-83 has been for `aarch64` only. The only significant difference to `ppc64le` I could find is that `aarch64` does not have https://github.com/CentOS/sig-cloud-instance-build/commit/2892c17fa8a520e58c3f42cd56587863fe675670#diff-574fa022ad7d0be565da139e942a15cbR92 , i.e., it has `override_install_langs=en_US.UTF-8` in `/etc/yum.conf` instead of `en_US.utf8`. If that is the culprit, we should change that (and upstream it, of course).

### DIFF
--- a/.circleci/test_docker_container.sh
+++ b/.circleci/test_docker_container.sh
@@ -2,6 +2,12 @@
 
 set -xeuo pipefail
 
+# test for working en_US.UTF-8 locale
+locale -a | grep -i 'en_US.UTF.\?8'
+[ "$( LC_ALL=en_US.UTF-8 sh -euc : 2>&1 )" = "" ]
+# make sure the above fails for non-existent locale
+[ ! "$( LC_ALL=badlocale sh -euc : 2>&1 )" = "" ]
+
 # check that /opt/conda has correct permissions
 touch /opt/conda/bin/test_conda_forge
 
--- a/.circleci/test_docker_container.sh
+++ b/.circleci/test_docker_container.sh
@@ -2,11 +2,11 @@
 
 set -xeuo pipefail
 
-# test for working en_US.UTF-8 locale
+echo 'Validating correctly configured en_US.UTF-8 locale...'
 locale -a | grep -i 'en_US.UTF.\?8'
-[ "$( LC_ALL=en_US.UTF-8 sh -euc : 2>&1 )" = "" ]
+[ "$( LC_ALL=en_US.UTF-8 sh -c : 2>&1 )" = "" ]
 # make sure the above fails for non-existent locale
-[ ! "$( LC_ALL=badlocale sh -euc : 2>&1 )" = "" ]
+[ ! "$( LC_ALL=badlocale sh -c : 2>&1 )" = "" ]
 
 # check that /opt/conda has correct permissions
 touch /opt/conda/bin/test_conda_forge
--- a/linux-anvil-cuda/Dockerfile
+++ b/linux-anvil-cuda/Dockerfile
@@ -36,7 +36,8 @@ RUN yum update -y && \
         localedef -i en_US -f UTF-8 en_US.UTF-8 && \
         : > /usr/lib/locale/locale-archive.tmpl ; \
     fi && \
-    yum clean all
+    yum clean all && \
+    rm -rf /var/cache/yum/*
 
 # Run common commands
 COPY scripts/run_commands /opt/docker/bin/run_commands
--- a/linux-anvil-cuda/Dockerfile
+++ b/linux-anvil-cuda/Dockerfile
@@ -20,14 +20,6 @@ ADD http://worldclockapi.com/api/json/utc/now /opt/docker/etc/timestamp
 # Resolves a nasty NOKEY warning that appears when using yum.
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-6
 
-# Fix up issues with locales because the default images have these minimized
-# to the point of not being properly functional
-# See: https://github.com/CentOS/sig-cloud-instance-images/issues/71
-RUN yum update -y && \
-    yum reinstall -y glibc-common && \
-    yum clean all
-RUN localedef -i en_US -f UTF-8 en_US.UTF-8
-
 # Remove preinclude system compilers
 RUN rpm -e --nodeps --verbose gcc gcc-c++
 
--- a/linux-anvil-cuda/Dockerfile
+++ b/linux-anvil-cuda/Dockerfile
@@ -24,20 +24,14 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-6
 RUN rpm -e --nodeps --verbose gcc gcc-c++
 
 # Install basic requirements.
+COPY scripts/yum_clean_all /opt/docker/bin/
 RUN yum update -y && \
     yum install -y \
         bzip2 \
         sudo \
         tar \
         which && \
-    : '(centos:6 only) If glibc-common is updated, we have to manually purge non-en_US locales.' && \
-    if locale -a | grep nds_DE ; then \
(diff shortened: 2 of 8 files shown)


.circleci/test_docker_container.sh
```diff
@@ -2,6 +2,12 @@
 
 set -xeuo pipefail
 
+# test for working en_US.UTF-8 locale
+locale -a | grep -i 'en_US.UTF.\?8'
+[ "$( LC_ALL=en_US.UTF-8 sh -euc : 2>&1 )" = "" ]
```
mbargull: This would spit out
```
/usr/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory
```
on `stderr` if the locale wouldn't work. If it works, we just get no output and can test for that.

mbargull:
> The only significant difference to `ppc64le` I could find is that `aarch64` does not have https://github.com/CentOS/sig-cloud-instance-build/commit/2892c17fa8a520e58c3f42cd56587863fe675670#diff-574fa022ad7d0be565da139e942a15cbR92 , i.e., it has `override_install_langs=en_US.UTF-8` in `/etc/yum.conf` instead of `en_US.utf8`. If that is the culprit, we should change that (and upstream it, of course).

Yep, that was is. All fixed now. I'll open an upstream PR.

mbargull:
> I'll open an upstream PR.

xref: https://github.com/CentOS/sig-cloud-instance-build/pull/154

linux-anvil-cuda/Dockerfile
```diff
@@ -38,7 +30,14 @@ RUN yum update -y && \
         sudo \
         tar \
         which && \
-    yum clean all
+    : '(centos:6 only) If glibc-common is updated, we have to manually purge non-en_US locales.' && \
```
mbargull: Docker doesn't like real comments as part of a multi-line `RUN` so we have to work around that if we want them.
(Ab-)using the null utility (special built-in) `:` for this is a little hacky, TBH, but it *really* doesn't do any harm, so it is perfectly fine in my book.

mbargull:
This is good to go, please take a look.
What this does:
- add tests for working locale
- properly fix the `aarch64` locale issue (write `override_install_langs=en_US.utf8` to `/etc/yum.conf`)
- properly rebuilt locale archive on `centos:6` (removes `~ 90 MiB`)
- clean caches more thoroughly (i.e., `conda clean -i` and `rm -rf /var/cache/yum/*`)
- add missing GPG keys for RPM on `aarch64` and `ppc64le`
- show `docker history` when testing (can be used to inspect layer sizes if `--squash` is (temporarily) disabled)

.circleci/test_docker_container.sh
```diff
@@ -2,6 +2,12 @@
...
+# make sure the above fails for non-existent locale
+[ ! "$( LC_ALL=badlocale sh -euc : 2>&1 )" = "" ]
```
hmaarrfk: nice

hmaarrfk:
Looks pretty good, but I'm no expert :/

mbargull:
Oh, and apart from testing/fixing the locale, this also shrinks the (squashed, uncompressed) `-comp7` image by `~ 30 %` (total `1.28 GB` vs `1.82 GB` total or `1.08 GB` vs `1.63 GB` for the layers we add).

mbargull:
Thanks for the review @hmaarrfk!

.circleci/test_docker_container.sh
```diff
@@ -2,6 +2,12 @@
 
 set -xeuo pipefail
 
+# test for working en_US.UTF-8 locale
```
mariusvniekerk: Can we add some echo messages here so that is is clear in the logs what is actually going on here?  Mostly just a line saying validate utf8 locale configured correctly.

.circleci/test_docker_container.sh
```diff
@@ -2,6 +2,12 @@
 
 set -xeuo pipefail
 
+echo 'Validating correctly configured en_US.UTF-8 locale...'
```
mbargull: @mariusvniekerk, like so?

linux-anvil-cuda/Dockerfile
jakirkham: Could we move this out into a bash script that we add and run to all `Dockerfile`s? This will ensure we only need to update and maintain this in one spot (as opposed to every Docker image), which will keep things maintainable here. Also we can format this more nicely without having to deal with Docker's syntax oddities. 🙂

mbargull: Sure, I had the same thing in mind. I was just in the "meh, I'll likely refactor stuff later anyway when proposing multi-staged builds, so whatever" state :wink: .

scripts/yum_clean_all
```diff
@@ -0,0 +1,12 @@
+#! /bin/sh -eu
+
+# (centos:6 only) If glibc-common is updated, we have to manually purge non-en_US locales.
+if grep -F 'CentOS release 6' /etc/centos-release ; then
+    if locale -a | grep nds_DE ; then
+        cp /usr/lib/locale/locale-archive /usr/lib/locale/locale-archive.tmpl
```
mbargull: (`cp` instead of `mv` to avoid writing to the mmapped file later.)

linux-anvil-cuda/Dockerfile
mbargull: Now it uses a common script in all images so that the `rm -rf /var/cache/yum/*` addition is only in one place, too.

scripts/yum_clean_all
jakirkham: Should we make that a comment in the script?

mbargull: Nah, not really important. The only change I expect to happen to this code is to be removed whenever this isn't needed anymore, e.g., when we use Centos 7 everywhere.
I just added the comment for my future self if I ever wondered why I changed that.

mbargull:
This would be ready to merge. However, Travis didn't run somehow... Any ideas?

mbargull:
Travis is running. But it's not listed as one of the "checks" needed. Strange.

linux-anvil-ppc64le/Dockerfile
```diff
@@ -17,14 +17,14 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
...
+COPY scripts/yum_clean_all /opt/docker/bin/
 RUN yum update -y && \
     yum install -y \
         bzip2 \
         sudo \
         tar \
         which && \
-    yum clean all && \
-    rm -rf /var/cache/yum/*
+    /opt/docker/bin/yum_clean_all
```
jakirkham: Maybe we should just move all of this into that script.

mbargull: Yes, this and other parts can/should be consolidated. But let's do further cleanup.maintenance work later, not in this PR.

mbargull:
Travis build finished.... and is listed as "Pending — The Travis CI build is in progress"... right... Well, ready to merge from my side.

mariusvniekerk:
If you really want the full green orb treatment you may want to restart the Travis build. I don't care too much about that :)

mbargull:
> the full green orb treatment 

Oh, I like many kinds of green hues -- olive green is a nice one; all good ;).
I was just waiting for John's nod. @jakirkham, you okay with the current state and postponing further cleanup and such?

PR #109, opened 2019-07-10
Use qemu 3.1.0-3

mbargull:
@mariusvniekerk in https://github.com/conda-forge/conda-build-feedstock/pull/101#issuecomment-510133675 :
> I'm not entirely sure what the build errors are on the emulated platforms. It suggests that there may be some issues with how qemu operates since we moved to 4.0.0-5 Should we roll that back to the last good 3.x release?

@mbargull in https://github.com/conda-forge/conda-build-feedstock/pull/101#issuecomment-510181473 :
> Looks like @mariusvniekerk is spot-on:
> ```sh
> >>> docker run --rm -it --entrypoint='' mbargull/condaforge-linux-anvil-aarch64:qemu-3.1.0-3 /opt/conda> /bin/python -c 'import os;os.listxattr(".");print("all good")'
> all good
> >>> docker run --rm -it --entrypoint='' mbargull/condaforge-linux-anvil-aarch64:qemu-4.0.0-5 /opt/conda> /bin/python -c 'import os;os.listxattr(".");print("all good")'
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
> OSError: [Errno 38] Function not implemented: '.'
> ```

(cc @jjhelmus)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,9 +48,9 @@ before_install:
       if [ "$(uname -m)" = "x86_64" ]; then
           docker run --rm --privileged multiarch/qemu-user-static:register --reset
       fi
-      export QEMU_STATIC_VERSION=v4.0.0-5
-      qemu_ppc64le_sha256=8a62c08b1bf8f65abc9c34bd8d7d31015f16e27c2d78b62367e7b01d5436005b
-      qemu_aarch64_sha256=010b188f02b3b0269ee1618840b058ff05c6bace18421f7db16bee2c8a4a5d13
+      export QEMU_STATIC_VERSION=v3.1.0-3
+      qemu_ppc64le_sha256=d018b96e20f7aefbc50e6ba93b6cabfd53490cdf1c88b02e7d66716fa09a7a17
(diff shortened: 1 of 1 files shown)


mbargull:
Getting 503 errors for `ADD http://worldclockapi.com/api/json/utc/now /opt/docker/etc/timestamp` on `master`... Should we just wait/retry or is this more urgent for you? Do you have an alternative service/URL?

mbargull:
Opened bug report at https://bugzilla.redhat.com/show_bug.cgi?id=1731756 .
Interestingly enough, the QEMU 4.0.0 from Alpine Linux does not have this problem. But that one isn't usable for us since it has the `ppc64le` bug that caused trouble with OpenSSL in https://github.com/conda-forge/docker-images/pull/106 .

mbargull:
Fedora build is fixed for QEMU 4.1.0 RC2 onward.
Acc. to https://wiki.qemu.org/Planning/4.1 4.1.0 will be released on 2019-08-06 or 2019-08-13 .
I would say we can give QEMU 4 another spin once that is released and built.

jjhelmus:
Thanks for looking into this @mbargull

PR #132, opened 2020-04-18
Add Dockerfile for win64 debugging

xhochy:


wolfv:
Awesome docker file, I just tested it locally. The entrypoint syntax doesn't seem to work on the stable docker, is that possible? had to condense everything into one line.

Also, I think launching into bash from conda's git package would be super nice :)

xhochy:
> Awesome docker file, I just tested it locally. The entrypoint syntax doesn't seem to work on the stable docker, is that possible? had to condense everything into one line.

Yes, I'm currently using custom per-package `ENTRYPOINT` and this got broken as I cleaned up this image.

> Also, I think launching into bash from conda's git package would be super nice :)

Actually I only switch into bash when I edit things, otherwise I need `cmd` to call the base scripts like `bld.bat` for failed conda builds.

wolfv:
yep, I think switching to bash as default is not such a great idea after all.

PR #145, opened 2020-07-28
Drop `armv7l` images

jakirkham:
As it appears we are not using this image or publishing packages for `armv7l`, go ahead and drop the `linux-anvil-armv7l` image.

cc @conda-forge/core


jjhelmus:
LGTM, if anyone wants to start 32-bit ARM builds on conda-forge this can always be re-enabled.

mbargull:
> if anyone wants to start 32-bit ARM builds on conda-forge

Psst! Don't give people ideas (they might go to a certain "Helmus" for help then) ;P.

PR #155, opened 2020-09-22
cuda 11

scopatz:
Adds a cuda11 image


scopatz:
For reference, I tried building this locally and it worked.

jakirkham:
Thanks @scopatz! Have been meaning to do this for a while. Sorry for the delay

How do we handle the CentOS 6/7 naming convention in other cases?

linux-anvil-cos7-cuda/Dockerfile
```diff
@@ -0,0 +1,111 @@
+ARG CUDA_VER
+
+FROM nvidia/cuda:${CUDA_VER}-devel-centos7
```
jakirkham, line 3: I wonder if we can template this with a build `ARG` so that we can reuse the existing image a bit. Just trying to figure out how we can keep the maintenance burden low.

scopatz: yeah, I could try to template on centos version too, if you would like

jakirkham: If it's not too difficult, otherwise we can skip it. I'm also happy to look as well :)

scopatz: yeah, working on it right now

jakirkham: Thanks Anthony! Please let me know if you need a hand.

scopatz: Will do thanks John!

scopatz:
Probably ready for another round of review @jakirkham

scopatz:
Just to see if I am headed in the right direction.  Note this issue: https://github.com/docker/distribution/issues/2459#issuecomment-348848758

jakirkham:
Thanks Anthony! Yeah this seems sensible to me. Let's see what CI thinks 🙂

jakirkham (approved):

PR #158, opened 2020-10-16
CentOS 8 for ARM

scopatz:
As per title. CC @isuruf @jakirkham

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,12 @@ matrix:
 
     - os: linux
       env:
+        - DOCKERIMAGE=linux-anvil-aarch64
+        - DOCKERTAG=8
+        - CENTOS_VER=8
+
+    - os: linux
+      env:
         - DOCKERIMAGE=linux-anvil-cuda
         - DOCKERTAG=9.2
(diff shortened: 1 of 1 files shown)


scopatz:
For https://github.com/conda-forge/cudatoolkit-feedstock/pull/4

jakirkham:
I think we will want an NVIDIA based image like we have on `x86_64` and `ppc64le` so that we can get the dev tools as well. The image name is [`nvidia/cuda-arm64`]( https://hub.docker.com/r/nvidia/cuda-arm64/tags ). These were only added recently. So we only have CUDA 11.0+ support.

scopatz:
OK, yeah, I think I am less interested in doing that right now. I'll leave this here for others to take a look or take over if they want.

jakirkham:
cc @JasonAtNvidia (for vis)

jakirkham:
The underlying use case was addressed by PR ( https://github.com/conda-forge/docker-images/pull/189 )

jakirkham:
Thanks for the work here Anthony! 😄

PR #171, opened 2021-01-19
ENH Add CUDA 11.2 support and update nvidia/cuda docker tags

mike-wendt:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

- Adds CUDA 11.2 support
- Updates docker tags used in `nvidia/cuda` which now include the [patch version](https://hub.docker.com/r/nvidia/cuda) for versions 11.0+

**NOTE:** This changes the use of `CUDA_VER` for CUDA `11.x`; open on suggestions as to whether the [documentation change](https://github.com/conda-forge/docker-images/pull/171/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R27-R29) is enough or there needs to be a new variable for the `nvidia/cuda` tag with major-minor-patch instead of using `CUDA_VER` as it may have unintended impacts.

cc @jakirkham

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -88,6 +88,11 @@ jobs:
           CUDA_VER: "11.1"
           CENTOS_VER: "7"
 
+        - DOCKERIMAGE: linux-anvil-cuda
+          DOCKERTAG: "11.2"
+          CUDA_VER: "11.2.0"
+          CENTOS_VER: "7"
+
         - DOCKERIMAGE: linux-anvil-ppc64le-cuda
           DOCKERTAG: "9.2"
           CUDA_VER: "9.2"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,12 +80,12 @@ jobs:
 
         - DOCKERIMAGE: linux-anvil-cuda
           DOCKERTAG: "11.0"
-          CUDA_VER: "11.0"
+          CUDA_VER: "11.0.3"
           CENTOS_VER: "7"
 
         - DOCKERIMAGE: linux-anvil-cuda
           DOCKERTAG: "11.1"
-          CUDA_VER: "11.1"
+          CUDA_VER: "11.1.1"
           CENTOS_VER: "7"
 
         - DOCKERIMAGE: linux-anvil-cuda
--- a/README.md
+++ b/README.md
@@ -24,7 +24,9 @@ docker build --rm --build-arg CENTOS_VER=6 --build-arg CUDA_VER=10.2 -f linux-an
 ## Environment variables
 
 * `$CUDA_VER`: This is the cuda & cudatoolkit version that will be used. The
-  value of this variable should be in major-minor for, e.g. `9.2`.
+  value of this variable should be in major-minor for, e.g. `9.2` for versions
+  `9.x` and `10.x`. For versions `11.x` the variable should be in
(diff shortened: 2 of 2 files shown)


mike-wendt:
I see failures as `CUDA_VER is used for `cudatoolkit` version. One is for 11.2 CTK which does not exist and needs this image to be built. What is the process for chicken and the egg problems like this? Do I need to build CTK 11.2 with another image to get this image built and then switch to using these images?

jakirkham:
Yeah I think we need to add a PR to [the `cudatoolkit` feedstock]( https://github.com/conda-forge/cudatoolkit-feedstock ) to add 11.2. That feedstock just uses the normal non-CUDA docker image. So shouldn't be blocked by this.

If we need this image out sooner for some reason, we can just skip downloading the `cudatoolkit` package on `11.2`. This is just done for caching purposes to cut down on package build time. So shouldn't have any adverse effects.

mike-wendt:
> Yeah I think we need to add a PR to [the `cudatoolkit` feedstock](https://github.com/conda-forge/cudatoolkit-feedstock) to add 11.2. That feedstock just uses the normal non-CUDA docker image. So shouldn't be blocked by this.
> 
> If we need this image out sooner for some reason, we can just skip downloading the `cudatoolkit` package on `11.2`. This is just done for caching purposes to cut down on package build time. So shouldn't have any adverse effects.

Yeah I'm working on one, will post shortly

jakirkham:
Ok happy to take a look whenever you are ready 🙂

mike-wendt:
> Ok happy to take a look whenever you are ready 🙂

@jakirkham are you good with my changes here to use `CUDA_VER` with `11.x` as major-minor-patch? This may break other things so the only other idea I had was to create a new variable in the config to be able to use the new tag format to pull the correct `nvidia/cuda` image but use `CUDA_VER` as is today with just major-minor

jakirkham:
That should be ok as long as NVIDIA's docker images use the same naming. Please see this line for context:

https://github.com/conda-forge/docker-images/blob/55a3a4bd0d94302d99687948b947d92a12ce768f/linux-anvil-cuda/Dockerfile#L4

Otherwise I don't see any issues.

Only other suggestion is we may want to add a similar line for the `ppc64le` image

leofang:
Is it possible to build cos7 images for CUDA 11.x as well? cuTENSOR could make use of it. 🙂

jakirkham:
CUDA 11.0+ is already CentOS 7. They don't support an earlier version of CentOS

leofang:
Ah I see, so the work is already done then, even better! 💪

jakirkham:
Yeah I think that image just got added so that there are CentOS 7 images for the older CUDA versions that also support CentOS 6 ( https://github.com/conda-forge/docker-images/pull/167 )

jaimergp:
The missing `cudatoolkit` is being provided here: https://github.com/conda-forge/cudatoolkit-feedstock/pull/46

jakirkham (approved):

leofang:
@jakirkham `ppc64le` images will be added in another PR, right?

jakirkham:
Yep if someone can grab that that would be great. Honestly still wrestling with Travis CI failures on the `cudatoolkit` feedstock. Errors seem to be about downloading the Docker image failing. So those packages are not yet published 😞

PR #198, opened 2021-10-25
Add CUDA `11.5` variants to CI axis

ajschmidt8:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

The `nvidia/cuda` version `11.5` images aren't out quite yet, but this PR prepares the `conda-forge` Docker images for when they do come out :slightly_smiling_face:.

PR #204, opened 2022-01-15
Add CUDA 11.6 images

jakirkham:


isuruf (approved):

jakirkham:
Thanks Isuru! 😄

@kkraus14 @jaimergp, could you please also take a look?

jaimergp (approved):

.github/workflows/ci.yaml
```diff
@@ -106,6 +106,11 @@ jobs:
           CUDA_VER: "11.5.1"
           CENTOS_VER: "8"
 
+        - DOCKERIMAGE: linux-anvil-cuda
+          DOCKERTAG: "11.6"
+          CUDA_VER: "11.6.0"
+          CENTOS_VER: "8"
```
kkraus14, line 112: Is there a reason we moved to CentOS 8 for CUDA 11.6 / 11.5 on x86? CentOS 7 is still supported as far as I can tell?

kkraus14 (changes requested):

.github/workflows/ci.yaml
jakirkham: Interesting I seem to recall CentOS 7 wasn't supported here before and recall us discussing this, but can't find that atm. Asking here ( https://github.com/conda-forge/docker-images/pull/198#discussion_r787116632 )

jakirkham: cc @ajschmidt8 @leofang

leofang: @jakirkham This is a question we should ask the image providers internally. If you search on the docket hub, there's no cos7 based image for CUDA 11.5.1, but there is one for 11.6.0.

jakirkham: Please see this issue ( https://gitlab.com/nvidia/container-images/cuda/-/issues/145 )

kkraus14: So should we use CentOS 7 for CUDA 11.6?

jakirkham: Checked with others and CentOS 7 should be supported by both. We just lack an image for 11.5 atm.

Yeah switching this to CentOS 7 makes sense. Would also suggest switching the 11.5 images to CentOS 7 once available. Have raised issue ( https://github.com/conda-forge/docker-images/issues/205 ) on this point. Am also subscribed to the upstream issue.

jakirkham:
```suggestion
          CENTOS_VER: "7"
```

leofang (approved):

kkraus14 (approved):

jakirkham:
Thanks all! 😄

Issue #205, opened 2022-01-19
Move CUDA 11.5 images to CentOS 7


jakirkham:
When adding CUDA 11.5 images, CentOS 7 was not available, but CentOS 8 was. It looks like the images were missing simply due to [a QA issue]( https://gitlab.com/nvidia/container-images/cuda/-/issues/145 ). Once that is resolved we should switch the CUDA 11.5 images here to CentOS 7.

cc @isuruf @jaimergp @kkraus14 @leofang (for vis)

jakirkham:
Now that these have been pushed, resolving in PR ( https://github.com/conda-forge/docker-images/pull/207 )

PR #207, opened 2022-01-25
Use CentOS 7 for CUDA 11.5 images

jakirkham:
Fixes https://github.com/conda-forge/docker-images/issues/205

Previously only CentOS 8 images were available for CUDA 11.5. So we used those. However we would prefer to use CentOS 7 images where possible (and do for other CUDA versions on `x86_64`). Now that CentOS 7 images are available for CUDA 11.5, switch to those.


kkraus14 (approved):

kkraus14:
Thanks @jakirkham!

Issue #184, opened 2021-06-03
`/usr/local/cuda/bin` is not added to `$PATH`


leofang:
It appears that #181 did not fix the problem, see https://github.com/conda-forge/cutensor-feedstock/pull/12#issuecomment-853318652.

leofang:
Looks like the aarch64 images have the same problem (https://github.com/conda-forge/docker-images/pull/189).

cc: @jakirkham (for awareness)

jakirkham:
For the PPC images, PR ( https://github.com/conda-forge/docker-images/pull/181 ) added this

https://github.com/conda-forge/docker-images/blob/91576a9507829f2a28b2ce5c5d7f39d59afb22b9/linux-anvil-ppc64le-cuda/entrypoint_source#L1-L2

<hr>

That said, ARM images currently lack those lines

jakirkham:
cc @jaimergp (in case you have thoughts on this as well)

jakirkham:
Actually I think this has to do without how we run with QEMU. Notice with [`x86_64` we add `-it` to `docker run`]( https://github.com/conda-forge/cutensor-feedstock/blob/6221d580efcadfd89da65a72a265861f1aaadc89/.scripts/run_docker_build.sh#L69-L71 ), but when using [QEMU we don't pass the `CI` environment variable]( https://github.com/conda-forge/cutensor-feedstock/blob/150be5793bca80ec8f87ee95fadd95aeb24cd1dc/.azure-pipelines/azure-pipelines-linux.yml#L42 ). So that check likely fails resulting in `-it` not being passed to `docker run`

jakirkham:
I'm wondering if this has something to do with how we are running QEMU. Normally when calling `docker run`, we add [`-it`]( https://github.com/conda-forge/cutensor-feedstock/blob/150be5793bca80ec8f87ee95fadd95aeb24cd1dc/.scripts/run_docker_build.sh#L69-L71 ), which should spin up a pseudo-TTY in interactive mode. When that wasn't done in the past, the source activation wouldn't work correctly.

jaimergp:
That's weird indeed and would explain why we still needed to manually export CUDA_PATH on the ppc builds for OpenMM, iirc.

jakirkham:
cc @mariusvniekerk (in case you have insights on how we might get this working with QEMU 🙂)

jakirkham:
[...]

PR #210, opened 2022-04-06
Copy `entrypoint_source` next to `Dockerfile`

jakirkham:
Fixes https://github.com/conda-forge/docker-images/issues/184

As this was copying the `entrypoint_source` files from the non-CUDA Docker images, it was only activating Conda and not adding CUDA executables to the `$PATH`.


jakirkham:
Requires PR ( https://github.com/conda-forge/docker-images/pull/211 ) to fix CI failures

jakirkham:
Merging to get these images fixed so we don't need to resort to hacks in feedstocks.

Issue #208, opened 2022-02-24
CUDA 11.0+ PPC & ARM builds failing


jakirkham:
Seeing the following failure on the CUDA 11.0+ images for PPC & ARM. Snippet taken from [one CI build]( https://github.com/conda-forge/docker-images/runs/5262962163?check_suite_focus=true#step:6:124 ) below:

```
Step 17/26 : RUN yum update -y --disablerepo=cuda &&     yum install -y         bzip2         sudo         tar         which         &&     /opt/docker/bin/yum_clean_all
Warning: rning] The requested image's platform (linux/ppc64le) does not match the detected host platform (linux/amd64) and no specific platform was requested
 ---> Running in 484977595ef4
CentOS Linux 8 - AppStream                       84  B/s |  38  B     00:00    
Failed to set locale, defaulting to C.UTF-8
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
The command '/bin/sh -c yum update -y --disablerepo=cuda &&     yum install -y         bzip2         sudo         tar         which         &&     /opt/docker/bin/yum_clean_all' returned a non-zero code: 1
```

Note the first build that passes is [this one from 25 days ago]( https://github.com/conda-forge/docker-images/actions/runs/1767881750 ).

jaimergp:
Timeline matches this [post](https://forums.centos.org/viewtopic.php?f=54&t=78708). COS8 is EOL :(

jakirkham:
Thanks Jaime 🙂 Makes sense

Is there a way to ignore the errors? Would be good to keep these functional (even though CentOS has moved on)

jaimergp:
I believe we need to change the repo urls to the "vault", as explained in this [post](https://forums.centos.org/viewtopic.php?p=330798#p330798):

```
sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
yum clean all
yum update -y
```

That will allow us to stay up-to-date (up to Dec 31st 2021). It might not last long, though.

jakirkham:
Yeah was looking at that

jakirkham:
A related question is how the `nvidia` base images may change due to this

[...]

PR #211, opened 2022-04-06
Fallback to CentOS Vault for CentOS 8 support

jakirkham:
Fixes https://github.com/conda-forge/docker-images/issues/208

Use CentOS Vault for `yum` to provide continued CentOS 8 support. Without this `yum` errors and the images won't build.

### DIFF
--- a/linux-anvil-aarch64-cuda/Dockerfile
+++ b/linux-anvil-aarch64-cuda/Dockerfile
@@ -45,6 +45,14 @@ RUN if [ "$CENTOS_VER" -le "7" ]; then \
 # Remove preinclude system compilers
 RUN rpm -e --nodeps --verbose gcc gcc-c++
 
+# Fallback to CentOS vault for CentOS 8 support.
+RUN if [ "$CENTOS_VER" -eq "8" ]; then \
+        find /etc/yum.repos.d/ -name "CentOS-*.repo" -exec \
+             sed -i 's/mirrorlist/#mirrorlist/g;s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' {} \; && \
+        yum update -y && \
+        yum clean all; \
+    fi
+
 # Install basic requirements.
 COPY scripts/yum_clean_all /opt/docker/bin/
 RUN yum update -y --disablerepo=cuda && \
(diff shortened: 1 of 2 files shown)


hmaarrfk (approved):

jakirkham:
Thanks Mark! 😄

Merging to get these building again

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -45,6 +45,14 @@ RUN if [ "$CENTOS_VER" -le "7" ]; then \
 # Remove preinclude system compilers
 RUN rpm -e --nodeps --verbose gcc gcc-c++
 
+# Fallback to CentOS vault for CentOS 8 support.
+RUN if [ "$CENTOS_VER" -eq "8" ]; then \
+        find /etc/yum.repos.d/ -name "CentOS-*.repo" -exec \
+             sed -i 's/mirrorlist/#mirrorlist/g;s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' {} \; && \
+        yum update -y && \
```
jakirkham, line 52: Needed to ensure CUDA packages are not updated unintentionally. This flag was used in the command below as well

```suggestion
        yum update -y --disablerepo=cuda && \
```

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -45,6 +45,14 @@ RUN if [ "$CENTOS_VER" -le "7" ]; then \
...
+        yum clean all; \
```
jakirkham, line 53: Elsewhere a custom script is being used for `yum clean all`. This is primarily needed for CentOS 6, which isn't relevant here. However it does all delete files in the cache, which would be useful here. Plus this script is being used elsewhere. So this should be added here.

```suggestion
        yum_clean_all; \
```

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -45,6 +45,14 @@ RUN if [ "$CENTOS_VER" -le "7" ]; then \
...
+    fi
+
 # Install basic requirements.
 COPY scripts/yum_clean_all /opt/docker/bin/
```
jakirkham, line 57: This needs to be moved higher up to allow it to be used in the command above.

jakirkham (commented):
Caught a few issues noted below. These are in both `Dockerfile`s, but only commented on one here for simplicity. Should add these have all since been fixed in PR ( https://github.com/conda-forge/docker-images/pull/212 ).

PR #214, opened 2022-05-19
Add CUDA 11.7 images

bdice:
Adds CI images for CUDA 11.7. Follows the example of #204.


jakirkham (approved):

.github/workflows/ci.yaml
```diff
@@ -166,6 +171,11 @@ jobs:
           CUDA_VER: "11.6.0"
           CENTOS_VER: "8"
 
+        - DOCKERIMAGE: linux-anvil-ppc64le-cuda
+          DOCKERTAG: "11.7"
+          CUDA_VER: "11.7.0"
+          CENTOS_VER: "8"
```
jakirkham, line 177: We may need to change this to use some other image (probably UBI) as CentOS 8 has been dropped

.github/workflows/ci.yaml
```diff
@@ -201,6 +211,11 @@ jobs:
           CUDA_VER: "11.6.0"
           CENTOS_VER: "8"
 
+        - DOCKERIMAGE: linux-anvil-aarch64-cuda
+          DOCKERTAG: "11.7"
+          CUDA_VER: "11.7.0"
+          CENTOS_VER: "8"
```
jakirkham, line 217: We may need to change this to use some other image (probably UBI) as CentOS 8 has been dropped

jakirkham (commented):
Noted a few images that might need to change below as CentOS 8 is no longer supported

.github/workflows/ci.yaml
bdice: Cool. I can tweak the template and variables accordingly, so that this is configurable instead of hardcoded.

jakirkham: Sure let's give that try. AIUI UBI is pretty close to RHEL so it should work well for us. Though will be interesting to see if any differences pop up

bdice: From what I can see on [DockerHub](https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=11.7.0-devel-), UBI is the only image with ppc64le support. Seems like the right choice.

bdice: @jakirkham For the moment, I only replaced the image names in `linux-anvil-ppc64le-cuda/Dockerfile` and `linux-anvil-aarch64-cuda/Dockerfile`. I don't think replacing the variable name `CENTOS_VER` with something more generic would be very easy, given its prevalence in this repo.

bdice: Note: it looks like `ubi8` images are available for ppc64le and aarch64 for CUDA versions 11.0.3 and newer.

bdice:
CI is failing with `error: /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial: import read failed(2).`

The ubi8 image doesn't have the same key names. It does have a key `/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release`.

I'm investigating this and will push a fix momentarily.

.github/workflows/ci.yaml
jakirkham: Think if `CENTOS_VER` is `8` that should be close enough to what we need (even though it isn't really CentOS)

Right we could use this for the other images as well. Maybe we should hold off on that for now

jakirkham:
Maybe we can do `/etc/pki/rpm-gpg/RPM-GPG-KEY-*` or similar?

bdice:
I verified locally that `/etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release` exists for both ubi7 and ubi8 images (so the `CENTOS_VER` conditional doesn't matter here). I'm going to see if CI passes/fails, and I'll try `/etc/pki/rpm-gpg/RPM-GPG-KEY-*` if it doesn't work.

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -35,12 +35,7 @@ ADD https://loripsum.net/api /opt/docker/etc/gibberish
 # Naming convention changed with cos8 - see:
 # * https://lists.centos.org/pipermail/centos-devel/2019-September/017847.html
 # * https://www.centos.org/keys/#project-keys
-RUN if [ "$CENTOS_VER" -le "7" ]; then \
-        rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-${CENTOS_VER} && \
-        rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-${CENTOS_VER}-aarch64; \
-    else \
-        rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial; \
-    fi
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
```
jakirkham, line 38: Probably need the same thing here

https://github.com/conda-forge/docker-images/blob/759e147952e4da0413dacc007e843871398354aa/linux-anvil-ppc64le-cuda/Dockerfile#L55-L60

Maybe we can update the comment above as well

jakirkham:
Was playing around with the UBI images locally. It seems `yum` is pretty limited. For example using [these requirements]( https://github.com/conda-forge/ucx-split-feedstock/blob/5b3a1d3014b428e26b7f94c7d6205e8d6b464b8b/recipe/yum_requirements.txt#L1-L2 ) saw the following

```bash
$ sudo yum install numactl-devel rdma-core-devel
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 0:00:40 ago on Thu May 19 20:11:47 2022.
No match for argument: numactl-devel
No match for argument: rdma-core-devel
Error: Unable to find a match: numactl-devel rdma-core-devel
```

linux-anvil-aarch64-cuda/Dockerfile
bdice: @jakirkham Can you offer guidance on what you'd put in the comment? I'm just not familiar enough to know what should be referenced here, if anything.

Oops, I meant to commit the same for ppc64le. Will fix.

Also, I verified that there are no architecture-specific signing keys for aarch64 like `/etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-${CENTOS_VER}-aarch64` in the previous code.

bdice:
@jakirkham Yes, it does seem that the software repositories for the `ubi8` image are very limited compared to CentOS. Another option would be Rocky Linux, which has `numactl-devel` and `rdma-core-devel` on x86_64 and aarch64. However, Rocky Linux is not currently offered for ppc64le.

bdice:
...unfortunately Rocky Linux builds are only available for CUDA 11.7. Reverting that idea.

bdice:
I'd appreciate others' thoughts on how to proceed at this point. The `ubi8` images might work on aarch64 and ppc64le to some extent (for CUDA >=11.0), but the lack of common packages makes them somewhat limited.

(In the meantime, cudatoolkit packages for 11.7 are still being deployed, so CI may fail for that reason.)

jakirkham:
I think `ubi8` is fine, but I don't think we should be replacing existing images with `ubi8` based ones. Instead think we should add this just for CUDA 11.7+ (11.6 and earlier would remain unchanged). Can see a few ways to do this:

1. Add conditionals in images to handle `ubi8` differences when relevant
2. Break out files for `ubi8` image into new directory and use those to build CUDA 11.7
3. Same as 2 + rename the conda-forge image for clarity

jakirkham:
cc @kkraus14 @jakirkham (in case either of you have thoughts here 🙂)

bdice:
> I think `ubi8` is fine, but I don't think we should be replacing existing images with `ubi8` based ones. Instead think we should add this just for CUDA 11.7+ (11.6 and earlier would remain unchanged). Can see a few ways to do this:
> 
>     1. Add conditionals/environment variables in images to handle `ubi8` differences when relevant
> 
>     2. Break out files for `ubi8` image into new directory and use those to build CUDA 11.7
> 
>     3. Same as 2 + rename the conda-forge image for clarity

Right, I agree those would all be viable paths forward. I also see another set of options where we do one of the above, but use Rocky Linux for aarch64 + CUDA 11.7. I don't know exactly how the lack of yum packages will affect feedstocks using the UBI images, but that seems important to consider. In my understanding the lack of yum packages in UBI images would prevent building the ucx feedstock on ppc64le with CUDA 11.7, for example. We may not need that at present due to CUDA Enhanced Compatibility, though?

jakirkham:
It's probably cleaner if we can do the same thing for `ppc64le` & `aarch64`

For the most part it shouldn't. CEC with CUDA 11.2 is used in nearly all cases (and that uses CentOS 8). There are some exceptions (like `ptxcompiler`), but that isn't the norm. Also not thinking of any case where both `yum` and the latest CUDA are needed

kkraus14:
> Also not thinking of any case where both `yum` and the latest CUDA are needed

Just because there aren't any packages in conda-forge with this today, doesn't mean there aren't packages out there that would potentially want to make it into conda-forge. CEC is more about being able to compile with a NEWER minor CUDA version and then be able to run on older toolkits than building on an older minor CUDA version and running on a new toolkit. There's often compiler bugfixes that can make a huge difference for projects.

What's the harm in using CentOS 8 for the pre 11.7 images and Rocky Linux 8 for 11.7+ for now?

All of this would no longer be a problem if NVIDIA would accelerate upstreaming the `cuda-*` packages into conda-forge...

jakirkham:
CentOS 8 images for CUDA 11.7 don't exist as NVIDIA discontinued them

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG CUDA_VER
 ARG CENTOS_VER
-FROM --platform=linux/arm64 nvidia/cuda:${CUDA_VER}-devel-centos${CENTOS_VER}
+FROM --platform=linux/arm64 nvidia/cuda:${CUDA_VER}-devel-${CENTOS_VER}
```
bdice, line 3: I've changed the meaning of `CENTOS_VER` to be more than an integer here, only for aarch64 and ppc64le (`"centos7"` or `"ubi8"` instead of just `"8"`). This is primarily a short-term fix to see if CI passes. I would like to know if others think this is too hacky, and if so, suggestions for what to do instead (introduce another variable named `IMAGE_BASE`, perhaps?).

bdice: @jakirkham @kkraus14 @jaimergp CI is passing now, with one caveat that the meaning of `CENTOS_VER` has changed for the aarch64 and ppc64le CUDA images as mentioned above. If you have thoughts on how this should move forward, please let me know.

jakirkham: This seems ok. If we wanted to be clearer, maybe we could do `${DISTRO_NAME}${DISTRO_VER}`

linux-anvil-aarch64-cuda/Dockerfile
```diff
@@ -35,21 +35,24 @@ ADD https://loripsum.net/api /opt/docker/etc/gibberish
 # Naming convention changed with cos8 - see:
 # * https://lists.centos.org/pipermail/centos-devel/2019-September/017847.html
 # * https://www.centos.org/keys/#project-keys
-RUN if [ "$CENTOS_VER" -le "7" ]; then \
-        rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-${CENTOS_VER} && \
-        rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-SIG-${CENTOS_VER}-aarch64; \
-    else \
+RUN if [ "$CENTOS_VER" = "centos7" ]; then \
```
jaimergp, line 38: Is this equality check correct? I think we should use `==`  or `-eq` but I always need to check these on the command line, so please double check :D 

(Same comment for the remaining changes that involve an `if` clause).

linux-anvil-aarch64-cuda/Dockerfile
bdice: That sounds good to me! Should that be done in this PR or a follow-up? I would like to merge this and keep the scope constrained to CUDA 11.7, then open a separate PR to refactor that for all images.

linux-anvil-aarch64-cuda/Dockerfile
bdice: I googled this as well — you can use single equals for string comparison with the single [. The commands would also fail if these comparisons weren’t working.

Source: https://linuxize.com/post/how-to-compare-strings-in-bash/
> Use the = operator with the test [ command. Use the == operator with the [[ command for pattern matching.

jakirkham: Yeah both work. Shells are strange. Generally agree `-eq` is clearer than `=` because the latter typically makes one think of single assignment

linux-anvil-aarch64-cuda/Dockerfile
jakirkham: Given we are changing the meaning of the term in this PR, it seems changing the name would be appropriate in this PR. It could be refactored into an earlier PR that we merge first, but it should get handled in some way.

FWIW I started to do this here ( https://github.com/jakirkham/conda-forge-docker-images/commit/9dfc91f2759af5e3bbe9d628d2443b90937098a3 ), but am being pulled into other things so may not be able to finish it. Feel free to grab that as a starting point.

linux-anvil-aarch64-cuda/Dockerfile
jakirkham: Ah ok. Fair enough

linux-anvil-aarch64-cuda/Dockerfile
jakirkham: Submitted PR ( https://github.com/conda-forge/docker-images/pull/215 )

jakirkham: That's now in. So this can be rebased or merged with that

linux-anvil-ppc64le-cuda/Dockerfile
```diff
@@ -55,27 +55,29 @@ ADD https://loripsum.net/api /opt/docker/etc/gibberish
...
 # Fallback to CentOS vault for CentOS 8 support.
-RUN if [ "$DISTRO_VER" -eq "8" ]; then \
+RUN if [ "${DISTRO_NAME}${DISTRO_VER}" = "centos8" ]; then \
         find /etc/yum.repos.d/ -name "CentOS-*.repo" -exec \
              sed -i 's/mirrorlist/#mirrorlist/g;s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' {} \; && \
         yum update -y --disablerepo=cuda && \
         /opt/docker/bin/yum_clean_all; \
     fi
 
+# Remove preinclude system compilers
```
bdice, line 78: I moved this command down, to be consistent with the same change for aarch64 in #215.

linux-anvil-aarch64-cuda/Dockerfile
bdice: This update is now complete! I'll resolve this thread.

linux-anvil-ppc64le-cuda/Dockerfile
jakirkham: Yeah that makes. Didn't mean to have the change slip into that PR

That said, we should probably do this after all `yum update` commands

bdice: Done in 93799ede3e175a5ca9e3681e5b475fb7436bc053.

jakirkham (approved):

PR #234, opened 2023-05-15
Disable keep-alive job on forks

jakirkham:
Currently keep-alive job runs on any repo (including forks), which means it is pushing commits to those forks. This behavior is only needed on the main repo to ensure cron jobs continue to run (even if there hasn't been much activity on this repo). Forks don't have this need. So disable this job on forks.


jakirkham:
@jaimergp, could you please review?

jakirkham:
Thanks Jaime! 🙏

PR #235, opened 2023-05-31
[WIP] Add AlmaLinux 8 based image

jakirkham:
* Adds an AlmaLinux based image for use in conda-forge builds
* Starts by using version 8 (though makes it easy to update this version later)
* Does a multiarch build (should simplify handling different supported architectures)


beckermr:
We definitely want "8" or maybe conda_2_28 somewhere in the image name.

linux-anvil-alma/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+# Resolves a nasty NOKEY warning that appears when using yum.
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
+
+# Install basic requirements.
+COPY scripts/yum_clean_all /opt/docker/bin/
+RUN yum update -y && \
+    yum install -y \
+        bzip2 \
+        glibc-langpack-en \
```
beckermr, line 35: We used the minimal lang pack in the sysroot. I think that'd be better here since it includes more languages.

beckermr (changes requested):

linux-anvil-alma/Dockerfile
jakirkham: Sure we can do that. Something like this?

```suggestion
        glibc-minimal-langpack \
```

Initially was just trying to get us past the locale failures on CI, which this seems to have done. Unlike CentOS 7, this wasn't preinstalled for us (hence the CI failure from before). This seems to have fixed it 🙂

beckermr: Yep that should work!

beckermr (approved):

linux-anvil-alma/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+# Resolves a nasty NOKEY warning that appears when using yum.
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
+
+# Install basic requirements.
+COPY scripts/yum_clean_all /opt/docker/bin/
+RUN yum update -y && \
+    yum install -y \
+        bzip2 \
+        glibc-minimal-langpack \
```
jakirkham, line 35: It looks like the locale issues have returned unfortunately 😞

AFAICT `glibc-minimal-langpack` is already installed (though happy to list it if we like)

What about something like this?

```suggestion
        glibc-minimal-langpack \
        glibc-langpack-en \
```

beckermr: Ahhhhh! This means we likely need that in the sysroot too.

jakirkham: Do we want to file this as [a sysroot issue]( https://github.com/conda-forge/linux-sysroot-feedstock/issues/new )?

Maybe there is a more minimal package than this one. Just noting this one seemed to solve the locale problem previously

beckermr: Yea let's file it.

jakirkham: Ok filed issue ( https://github.com/conda-forge/linux-sysroot-feedstock/issues/48 ). Please feel free to add anything else there 🙂

jakirkham:
SGTM. Thanks for helping get this across the finish line Matt! 🙏

linux-anvil-alma/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+        conda-forge::binutils_linux-ppc64le \
+        conda-forge::gcc_impl_linux-ppc64le \
+        conda-forge::gcc_linux-ppc64le \
+        conda-forge::gfortran_impl_linux-ppc64le \
+        conda-forge::gfortran_linux-ppc64le \
+        conda-forge::gxx_impl_linux-ppc64le \
+        conda-forge::gxx_linux-ppc64le \
+        conda-forge::libgcc-ng \
+        conda-forge::libgfortran-ng \
+        conda-forge::libstdcxx-ng && \
```
isuruf, line 77: Why are you downloading all? This is going to be a very big image

linux-anvil-alma/Dockerfile
```diff
@@ -0,0 +1,93 @@
+ARG DISTRO_ARCH
```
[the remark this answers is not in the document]
isuruf: Please re-use [linux-anvil-comp7/Dockerfile](https://github.com/conda-forge/docker-images/tree/main/linux-anvil-comp7)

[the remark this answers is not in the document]
jakirkham: Because they are different OSes and versions, which resulted in subtle differences

Also when we run into issues specific to this distro & version, it will be helpful to have them siloed

Most of the things that can be reused already are through refactored scripts that all `Dockerfile`s use

Think it would be better to keep this distinct

linux-anvil-alma/Dockerfile
jakirkham: It was easier to lump them into one step. Plus in the cross-compilation case many of these get downloaded anyways

Though if someone wants to propose conditional logic, that would be reasonable

isuruf: Let's remove these lines. This cache does not work most of the time anyways. This downloads gcc 13 whereas the recipes uses gcc 12. For the past couple of years, it's very rare that this caching worked because we always had a higher gcc version available, but recipes used lower ones.

linux-anvil-alma/Dockerfile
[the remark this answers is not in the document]
isuruf: What are these subtle differences? It would good to have a diff to see what changed for easier review.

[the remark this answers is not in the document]
jakirkham: I'm sorry, but I don't plan to pick up this task

From the conda-forge meeting earlier it sound like others don't care or (in some cases don't want) this change

Would suggest converting this to an issue if we want to discuss further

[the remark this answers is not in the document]
ocefpaf: Folks, having it separate makes sense as well as having the diff from the previous one. What can we do here to solve this? Maybe we post a diff of the files here to document that? Do a 2-PR-commit thing overriding the old file, that won't be used and then move it to the dir here, so we document this diff in the git history?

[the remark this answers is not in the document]
beckermr: Here is the diff

```
0a1
> ARG DISTRO_ARCH
3c4
< FROM ${DISTRO_NAME}:${DISTRO_VER}
---
> FROM --platform=linux/${DISTRO_ARCH} ${DISTRO_NAME}:${DISTRO_VER}
6a8
> ARG DISTRO_ARCH
9c11,12
< ENV DISTRO_NAME=${DISTRO_NAME} \
---
> ENV DISTRO_ARCH=${DISTRO_ARCH} \
>     DISTRO_NAME=${DISTRO_NAME} \
25c28
< RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-${DISTRO_VER}
---
> RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
31a35,36
>         glibc-minimal-langpack \
>         glibc-langpack-en \
41,61d45
< # Download and cache new compiler packages.
< # Should speedup installation of them on CIs.
< RUN source /opt/conda/etc/profile.d/conda.sh && \
<     conda activate && \
<     conda create -n test --yes --quiet --download-only \
<         conda-forge::binutils_impl_linux-64 \
<         conda-forge::binutils_linux-64 \
<         conda-forge::gcc_impl_linux-64 \
<         conda-forge::gcc_linux-64 \
<         conda-forge::gfortran_impl_linux-64 \
<         conda-forge::gfortran_linux-64 \
<         conda-forge::gxx_impl_linux-64 \
<         conda-forge::gxx_linux-64 \
<         conda-forge::libgcc-ng \
<         conda-forge::libgfortran-ng \
<         conda-forge::libstdcxx-ng && \
<     conda remove --yes --quiet -n test --all && \
<     conda clean -tiy && \
<     chgrp -R lucky /opt/conda && \
<     chmod -R g=u /opt/conda
< 
65c49
< COPY linux-anvil-cos7-x86_64/entrypoint_source /opt/docker/bin/entrypoint_source
---
> COPY linux-anvil-alma/entrypoint_source /opt/docker/bin/entrypoint_source
```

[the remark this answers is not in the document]
beckermr: Entrypoint scripts are identical.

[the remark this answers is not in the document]
beckermr: I agree that different linux distros should be siloed FWIW. Future alma versions should likely use the same image under the assumption that they won't break or change too many things.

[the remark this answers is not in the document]
beckermr: Let's merge.

PR #236, opened 2023-06-08
Drop `defaults` channel

jakirkham:
To address some repodata issues seen in `defaults` affecting `mamba`, drop the `defaults` channel.

### DIFF
--- a/scripts/run_commands
+++ b/scripts/run_commands
@@ -2,7 +2,7 @@
 
 set -exo pipefail
 
-export additional_channel="--add channels defaults"
+export additional_channel=""
 
 export miniforge_arch="$(uname -m)"
 export miniforge_version="22.9.0-0"


jakirkham:
cc @conda-forge/core

beckermr (approved):

wolfv (approved):

wolfv:
Thanks @jakirkham I think this is better anyways nowadays!

jakirkham:
Looking at the history, this was added a few years back ( https://github.com/conda-forge/docker-images/pull/128 ). IIUC that preceded `conda-forge` being a full-contained channel

johnhany97:
Great - once this is released I'll revert the change in https://github.com/conda-forge/libsolv-feedstock/pull/67 to use Conda instead of mamba.

jakirkham:
All jobs, but one have passed. The one job is still building a CUDA images that uses emulation and is going kind of slow (maybe hung?). It's not one of images we use in our builds. Going to merge to get these images out. If there are still issues with that job, we can follow up separately

PR #239, opened 2023-08-21
Update to Miniforge 23.3.1-0

jaimergp:
The `run_commands` script was still pointing to an old version of Mambaforge (22.9). With https://github.com/conda-forge/miniforge/pull/277 now merged, it makes sense to standardize this across operating systems (macOS and Windows pull from latest Miniforge).

### DIFF
--- a/scripts/run_commands
+++ b/scripts/run_commands
@@ -5,15 +5,15 @@ set -exo pipefail
 export additional_channel=""
 
 export miniforge_arch="$(uname -m)"
-export miniforge_version="22.9.0-0"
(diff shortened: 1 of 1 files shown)


hmaarrfk:
should we also pin osx and windows to specific versions?

jaimergp:
Ideally yes, but that'll involve PRs in different repos, and a conversation about how we approach version locking, supply chain updates, etc 😬

jakirkham (approved):

jakirkham:
Thanks Jaime! 🙏

PR #240, opened 2023-08-22
Use `conda`'s `libmamba` `solver`

jakirkham:
Now that `conda-libmamba-solver` ships with the installer. Opt-in to using the `libmamba` `solver` with `conda`.

### DIFF
--- a/scripts/run_commands
+++ b/scripts/run_commands
@@ -41,6 +41,7 @@ source /opt/conda/etc/profile.d/conda.sh
 conda activate
 conda config --set show_channel_urls True
 conda config --set channel_priority strict
+conda config --set solver libmamba
 conda config ${additional_channel} --add channels conda-forge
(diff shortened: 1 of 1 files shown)


jakirkham:
We can make this change here, which may be valuable just to have in the Docker images

Though we can also make this change in `conda-forge-ci-setup` to cover all OSes: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/259

Or we can do both

jaimergp:
I'd say we stick to https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/259. The default will soon change anyway.

jakirkham:
Yeah guess the benefit of having it here also would be the Docker image is preconfigured (as opposing to adding a step) for users that just grab the Docker image. Though have no strong feelings (given this is an easy change to make)

PR #237, opened 2023-06-30
Refresh CUDA Docker images based on EOL

jakirkham:
As [noted at the previous conda-forge meeting]( https://github.com/conda-forge/conda-forge.github.io/blob/main/src/orga/minutes/2023-06-14.md ), some of these Docker images are going away. Please see [this notice]( https://gitlab.com/nvidia/container-images/cuda/-/issues/209#note_1452067531 ) and [the container support policy]( https://gitlab.com/nvidia/container-images/cuda/-/blob/master/doc/support-policy.md )

Have rebuilt all of the images currently include in this repo with [this GHA job]( https://github.com/conda-forge/docker-images/actions/runs/5422331357 ) to ensure we have the latest build.

Some of the images just require a minor version bump, which is relatively straightforward.

With some of the `arm64` & `ppc64le` images, they are now no longer available with CentOS 8. So have moved to UBI.

Most of these changes do not affect images that we are using to build packages in conda-forge. However there is one exception to this. The CUDA 10 images are going away. As we have already discussed dropping CUDA 10.x support in the same conda-forge meeting, this merely gives us a reason to pickup the pace. Refreshed PR ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1708 ) to help make that move

PR #243, opened 2023-11-10
Move rest of CUDA arch images to UBI 8

jakirkham:
The CUDA `aarch64le` & `ppc64le` CentOS 8 images are no longer available. So move them to UBI 8 based images, which are still around. This is what was done for the other CUDA images already ( https://github.com/conda-forge/docker-images/pull/237 ).


jakirkham:
This will fix the CI errors we are seeing with these images (for [example]( https://github.com/conda-forge/docker-images/actions/runs/6759584539/job/18372395006#step:6:47 ))

PR #265, opened 2024-05-07
Use keep alive in API mode in-workflow

jaimergp:
The keep alive v2 action allows to use the GH API instead of dummy commits. However, it comes with a few pitfalls 🤔

- The API call keeps a _workflow_ alive, not the entire repository. So we need to either add the step to every workflow to keep alive, or use the workflows setting. This is prone to break.
- We need to update the permissions in the workflow too.

Since we have a single cronjob CI, in this repo I'm merging the two workflows for simplicity and a cleaner git history. Let me know if you prefer a separate workflow file + dummy commits, though.


jakirkham:
Thanks Jaime! 🙏

Let's give this a try

jakirkham (approved):

PR #269, opened 2024-06-12
QEMU 8.2.4

h-vetinari:
Update QEMU version

### DIFF
--- a/download-qemu-static.sh
+++ b/download-qemu-static.sh
@@ -14,15 +14,16 @@ DEBIAN_FRONTEND=noninteractive \
     sudo apt-get install --yes --no-install-recommends \
     ca-certificates curl libarchive-tools
 
-version='8.1.3'
-build='1.fc39'
+version='8.2.4'
+build='1.fc40'
 for arch in aarch64 ppc64le s390x; do
     curl -sL \
         "https://kojipkgs.fedoraproject.org/packages/qemu/${version}/${build}/x86_64/qemu-user-static-${arch/ppc64le/ppc}-${version}-${build}.x86_64.rpm" |
         bsdtar -xf- --strip-components=3 ./usr/bin/qemu-${arch}-static
 done
+
 sha256sum --check << 'EOF'
-3b78154b2aa59f886b530826b5dd01be03ce3c4b29751a814d724650bcc526b2  qemu-aarch64-static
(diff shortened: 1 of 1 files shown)


download-qemu-static.sh
```diff
@@ -14,15 +14,20 @@ DEBIAN_FRONTEND=noninteractive \
     sudo apt-get install --yes --no-install-recommends \
     ca-certificates curl libarchive-tools
 
-version='8.1.3'
-build='1.fc39'
+# see https://gitlab.com/qemu-project/qemu/-/tags for versions;
+# we use the RPMs from https://kojipkgs.fedoraproject.org/packages/qemu;
+# do not use qemu builds from unrelesed fedora versions, compare `build`
+# vs. https://en.wikipedia.org/wiki/Fedora_Linux_release_history
```
mbargull, line 20: Thanks for adding the comment!
I agree that we'd normally want a bit more "stable" instead of "bleeding edge".
There can (rarely) be situations in which we encounter known-fixed bugs for which it might be more practical to temporarily switch to the latest builds available (emulating stuff can be complicated/messy...)
Let's rather use "avoid"/"prefer" here:
```suggestion
# avoid qemu builds from unreleased fedora versions, compare `build`
# vs. https://en.wikipedia.org/wiki/Fedora_Linux_release_history;
# prefer non-`.0` patch releases to try to avoid potential new regressions;
# if possible, check https://gitlab.com/qemu-project/qemu/-/issues
# for relevant issues in old vs new version;
```
Otherwise, LGTM!

h-vetinari: Thanks!

mbargull (approved):

download-qemu-static.sh
mbargull: And to you!

mbargull: Oh, and what's also good is to give the logs of the recent (and upcoming) version from https://wiki.qemu.org/ChangeLog/ a skim.
E.g., https://wiki.qemu.org/ChangeLog/9.0#Memory_backends and https://wiki.qemu.org/ChangeLog/9.0#Migration mentions fixed regressions from 8.2 .
These can give good hints in case we suddenly encounter failures after updating the QEMU version we're using.

PR #273, opened 2024-07-02
Reuse one `Dockerfile` for all CentOS 7 builds

jakirkham:
* Synchronize content across the different CentOS 7 architecture builds
* Generalize `Dockerfile` to across architectures
* Configure CI to build with one `Dockerfile` for CentOS 7 builds


jakirkham:
Thoughts @conda-forge/core? 🙂

linux-anvil-cos7/Dockerfile
```diff
@@ -1,24 +1,36 @@
...
+ENV DISTRO_ARCH=${DISTRO_ARCH} \
+    DISTRO_NAME=${DISTRO_NAME} \
     DISTRO_VER=${DISTRO_VER}
 
 # Set an encoding to make things work smoothly.
 ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+
+# bust the docker cache so that we always rerun the installs below
+ADD https://loripsum.net/api /opt/docker/etc/gibberish
```
xhochy, line 22: Any reason you moved the cache busting before the qemu binaries?

jakirkham: The original idea of busting the cache was to make sure the images were rebuilt from scratch from the beginning (particularly on services like DockerHub, which always use a cache to build)

So had moved this before other steps

Though if there are other considerations, would be interested to learn more about them

isuruf (approved):

jakirkham:
@xhochy did you have any more thoughts on this one? 🙂

xhochy (approved):

jakirkham:
Thank you both! 🙏

PR #281, opened 2024-09-25
Bump QEMU to 8.2.7 (missing 8.2.4 binaries)

jakirkham:
Recently the weekly cron jobs here [started failing]( https://github.com/conda-forge/docker-images/actions/runs/10757570540/job/29831758482 ). This appears related to in the QEMU preparation step

```
+ curl -sL https://kojipkgs.fedoraproject.org/packages/qemu/8.2.4/1.fc40/x86_64/qemu-user-static-aarch64-8.2.4-1.fc40.x86_64.rpm
+ bsdtar -xf- --strip-components=3 ./usr/bin/qemu-aarch64-static
bsdtar: Error opening archive: Unrecognized archive format
```

However when following the URL back, it appears upstream has [dropped the 8.2.4 binaries]( https://kojipkgs.fedoraproject.org/packages/qemu/8.2.4/ )

Screenshot 2024-09-24 at 6 38 34 PM

Also it appears there have been a few QEMU patch releases recently

To fix CI and update QEMU, this switches to the latest QEMU 8.2 patch release, which is 8.2.7

There is also a QEMU 9.0.0 and 9.1.0. However the goal here primarily is to fix CI builds, have stuck with the latest 8.2. Perhaps an upgrade to QEMU 9 can be evaluated separately

### DIFF
--- a/download-qemu-static.sh
+++ b/download-qemu-static.sh
@@ -21,7 +21,7 @@ DEBIAN_FRONTEND=noninteractive \
 # prefer non-`.0` patch releases to try to avoid potential new regressions;
 # if possible, check https://gitlab.com/qemu-project/qemu/-/issues
 # for relevant issues in old vs new version;
-version='8.2.4'
+version='8.2.7'
 build='1.fc40'
 for arch in aarch64 ppc64le s390x; do
     curl -sL \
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @conda-forge/core

xhochy (approved):

PR #291, opened 2024-10-29
Provide `linux-anvil-cuda` on x64 with RHEL 8, consolidate image names, distinguish distro baseline in tag

h-vetinari:
The other two architectures are already on UBI 8

For https://github.com/conda-forge/conda-forge.github.io/issues/1941 / https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6283

Check out the available [tags](https://hub.docker.com/r/nvidia/cuda/tags?name=11.8.0); the required `-devel-{distro}` is only available for ppc if we use ubi8, so might as well use that distro for all architectures.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,36 +41,36 @@ jobs:
             DISTRO_VER: "7"
             SHORT_DESCRIPTION: "conda-forge build image for CentOS 7 on aarch64"
 
-          - DOCKERIMAGE: linux-anvil-cuda
+          - DOCKERIMAGE: linux-anvil-cuda11.8
             DOCKERFILE: linux-anvil-cuda
-            DOCKERTAG: "11.8"
+            DOCKERTAG: "cos7"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "amd64"
             DISTRO_NAME: "centos"
             DISTRO_VER: "7"
             SHORT_DESCRIPTION: "conda-forge build image for CentOS 7 on x86_64 with CUDA 11.8"
 
-          - DOCKERIMAGE: linux-anvil-cuda
+          - DOCKERIMAGE: linux-anvil-cuda11.8
             DOCKERFILE: linux-anvil-cuda
-            DOCKERTAG: "11.8"
+            DOCKERTAG: "ubi8"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "amd64"
             DISTRO_NAME: "ubi"
             DISTRO_VER: "8"
             SHORT_DESCRIPTION: "conda-forge build image for UBI 8 on x86_64 with CUDA 11.8"
 
-          - DOCKERIMAGE: linux-anvil-ppc64le-cuda
+          - DOCKERIMAGE: linux-anvil-ppc64le-cuda11.8
             DOCKERFILE: linux-anvil-cuda
-            DOCKERTAG: "11.8"
+            DOCKERTAG: "ubi8"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "ppc64le"
             DISTRO_NAME: "ubi"
             DISTRO_VER: "8"
             SHORT_DESCRIPTION: "conda-forge build image for UBI 8 on ppc64le with CUDA 11.8"
 
-          - DOCKERIMAGE: linux-anvil-aarch64-cuda
+          - DOCKERIMAGE: linux-anvil-aarch64-cuda11.8
             DOCKERFILE: linux-anvil-cuda
-            DOCKERTAG: "11.8"
+            DOCKERTAG: "ubi8"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "arm64"
             DISTRO_NAME: "ubi"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,15 @@ jobs:
             DOCKERTAG: "11.8"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "amd64"
+            DISTRO_NAME: "centos"
+            DISTRO_VER: "7"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -23,32 +23,32 @@ jobs:
...
             DISTRO_NAME: "centos"
             DISTRO_VER: "7"
-            SHORT_DESCRIPTION: "conda-forge build image for Cent0S 7 on aarch64"
+            SHORT_DESCRIPTION: "conda-forge build image for CentOS 7 on aarch64"
 
           - DOCKERIMAGE: linux-anvil-cuda
             DOCKERFILE: linux-anvil-cuda
             DOCKERTAG: "11.8"
             CUDA_VER: "11.8.0"
             DISTRO_ARCH: "amd64"
```
jakirkham, line 48:
```suggestion
          - DOCKERIMAGE: linux-anvil-cuda
            DOCKERFILE: linux-anvil-cuda
            DOCKERTAG: "11.8"
            CUDA_VER: "11.8.0"
            DISTRO_ARCH: "amd64"
            DISTRO_NAME: "centos"
            DISTRO_VER: "7"
            SHORT_DESCRIPTION: "conda-forge build image for Cent0S 7 on x86_64 with CUDA"
```

.github/workflows/ci.yaml
```diff
@@ -23,32 +23,32 @@ jobs:
...
-            DISTRO_NAME: "centos"
-            DISTRO_VER: "7"
-            SHORT_DESCRIPTION: "conda-forge build image for Cent0S 7 on x86_64 with CUDA"
+            DISTRO_NAME: "ubi"
+            DISTRO_VER: "8"
+            SHORT_DESCRIPTION: "conda-forge build image for UBI 8 on x86_64 with CUDA 11.8"
```
jakirkham, line 51:
```suggestion

          - DOCKERIMAGE: linux-anvil-cuda
            DOCKERFILE: linux-anvil-ubi8-cuda
            DOCKERTAG: "11.8"
            CUDA_VER: "11.8.0"
            DISTRO_ARCH: "amd64"
            DISTRO_NAME: "ubi"
            DISTRO_VER: "8"
            SHORT_DESCRIPTION: "conda-forge build image for UBI 8 on x86_64 with CUDA 11.8"
```

jakirkham (commented):
Thanks Axel! 🙏

Generally this seems reasonable

In the past when upgrading from CentOS 6 to 7, we added new images that were named `linux-anvil-cos7-cuda` ( https://github.com/conda-forge/docker-images/pull/167 ). Am thinking it would make sense to do the same thing here and add `linux-anvil-ubi8-cuda`. Effectively this is just one more image (as the others are already on UBI 8)

Thinking this would minimize surprise for users as they would continue to get the same OS versions with the same image names (as they may have come to expect)

Admittedly CentOS 7 support is rolling off ( https://github.com/conda-forge/cuda-feedstock/issues/28 ). So we may end up dropping the old image at that time

h-vetinari:
Thanks for the review!

> In the past when upgrading from CentOS 6 to 7, we added new images that were named `linux-anvil-cos7-cuda` ( #167 ). Am thinking it would make sense to do the same thing here and add `linux-anvil-ubi8-cuda`.

For the alma and CUDA images, we've been moving to image tags per distro version, which I think is more elegant. I don't want to go back to including the distro in the image name (I don't mind putting it in tag though, if we really want to keep a cos7 variant selectable).

> Thinking this would minimize surprise for users as they would continue to get the same OS versions with the same image names (as they may have come to expect)

I was surprised to see that the image version (and distro) did not align for the cuda images between x64 resp. aarch/ppc, so I don't think there are many expectations of that sort (as the current situation is already inconsistent/confusing, and the image version doesn't matter in any practical sense since everything is based on the `sysroot` version). So I consider this a negligible risk, especially as all our infra will soon move to RHEL8-based images anyway.

PS. We've been waiting for ~2 weeks for a response from you in #287, please have a look at that PR as well. 🙏

h-vetinari:
@jakirkham, if you'd like to make the distro-version selectable for the CUDA 11.8 images, I have a proposal for how this could look like in #293; if we decide not to encode the distro version (which is completely fine IMO, as it's effectively irrelevant, c.f. https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6283), then this PR is ready.

h-vetinari:
This now incorporates the discussion from #291, in particular, the move of the distroname into the tag, in order to support choosing this where necessary (e.g. for feedstocks that do binary repackaging).

jakirkham:
Thanks Axel! 🙏

Just to clarify, the last comment is referring to issue ( https://github.com/conda-forge/docker-images/issues/293 )?

h-vetinari:
> Just to clarify, the last comment is referring to issue ( #293 )?

Yes, I messed up the xref, sorry. I meant this [comment](https://github.com/conda-forge/docker-images/issues/293#issuecomment-2474390077)

h-vetinari:
More specifically, the CUDA 11.8 portion of that comment. For the rest of the renames, my intention was to first progress on https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6626. I guess we could also do everything in one go, which would mean picking the additions from https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6687 into the other PR.

jakirkham:
All good. Just wanted to make sure I understood

Am happy with either approach

jakirkham (approved):

h-vetinari:
I hadn't pushed https://github.com/conda-forge/docker-images/pull/291/commits/63a890c3203dbe66a833e992c3e0e24f3aa0e369 until today (even though I had prepared it a while ago) - to be clear, it came after your approval even though it's shown chronologically as before.

However, given that the remaining changes strictly follow what was discussed in #293 (and in core), I think this is good to go.

Thanks for the review! 🙏

PR #299, opened 2024-12-12
Enable CRB repo on alma9, PowerTools on alma8

h-vetinari:
With the switch to alma8 & alma9, some of our CDTs are now split across separate repos. In particular, we need `PowerTools` on alma8, resp. `CRB` on alma9 (which is essentially the same thing but got renamed). See for example https://github.com/conda-forge/cdt-builds/commit/e118347a73c4927c902aff7b928fa1dfafa65782

As an example, the following failed:
```
+ /usr/bin/sudo -n yum install -y wget m4 help2man patch gcc-gfortran gcc-c++ rsync sed findutils make file strace
AlmaLinux 9 - AppStream                          30 MB/s | 9.2 MB     00:00    
AlmaLinux 9 - BaseOS                            3.2 MB/s | 4.5 MB     00:01    
AlmaLinux 9 - Extras                             86 kB/s |  13 kB     00:00    
No match for argument: help2man
```
That package is [available](https://almalinux.pkgs.org/9/almalinux-crb-x86_64/help2man-1.48.2-3.el9.noarch.rpm.html) under `CRB` on alma9, so enable this repo.

### DIFF
--- a/scripts/fix_rpm
+++ b/scripts/fix_rpm
@@ -18,15 +18,21 @@ elif [ "${DISTRO_NAME}${DISTRO_VER}" = "ubi8" ]; then
   rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
 elif [ "${DISTRO_NAME}${DISTRO_VER}" = "almalinux8" ]; then
   rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-AlmaLinux
+  ls -l /etc/yum.repos.d
+  cat /etc/yum.repos.d/AlmaLinux-PowerTools.repo
   # PowerTools repo not enabled by default
   sed -i '/^enabled=/d; /^\[/a\enabled=1' /etc/yum.repos.d/AlmaLinux-PowerTools.repo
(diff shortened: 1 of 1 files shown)


h-vetinari:
I'm fine for this PR to get squashed, if someone prefers that. I didn't squash the PR myself to show why I arrived at the solution here. Note that the following sed-command
https://github.com/conda-forge/docker-images/blob/7023391936aa2d4820987e5f291a5a051ab7ad14/scripts/fix_rpm#L7
is pretty gnarly to understand, and in particular, I couldn't find a way to chain another append-command afterwards.

I added some comments to explain a bit what's going on. Whereas the regex for cos7 unconditionally disables something (so adding `enable=0` everywhere is fine), here we only want to activate CRB resp. PowerTools, but without their sources or debuginfo flavours; hence we need two append commands.

jaimergp (approved):
This LGTM but I'd prefer if someone else can also take a look before merging because I'm not that in touch with this part of the stack.

h-vetinari:
Ping @conda-forge/core, would be nice if someone could review. :)

PR #300, opened 2024-12-22
Back to QEMU 8.x

h-vetinari:
The test suites in numpy and scipy show that QEMU 9 is massively slower than QEMU 8, and is causing the scipy test suite to hit the 6h timeout for roughly 4 out of 5 runs. The upgrade to v9 was made in the hopes of improving the testing situation in emulation, but this turned out to be a step back.

See also discussion in https://github.com/conda-forge/openblas-feedstock/issues/160


h-vetinari:
Merging this as a regression fix

PR #306, opened 2025-03-20
Add `.txt` to the time URL

jakirkham:
Appending `.txt` appears to work more reliably while giving the same result. So tweak the time URL accordingly.


jakirkham:
Lol that wasn't any better

Maybe it is time to drop this step: https://github.com/conda-forge/docker-images/pull/307

PR #314, opened 2025-09-18
Replace deprecated keepalive action with `gh api` call

### LINKED ISSUE #313: Keep-alive workflow disabled, resulted in no image rebuilds in a while
[Last run was three months ago](https://github.com/conda-forge/docker-images/actions/runs/15952228449). That workflow has been disabled ever since. I have manually re-enabled now, but we need to come up with a solution for this.

jaimergp:
Closes #313

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -219,5 +219,15 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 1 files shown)


wolfv:
Should we do the same for the conda-mirror job?

jaimergp:
Yep, we have a [couple more places](https://github.com/search?type=code&q=org%3Aconda-forge+OR+org%3Aregro+OR+org%3Achannel-mirrors+keepalive-workflow) with this deprecated action. I think I'll create a centralized action for us to reuse internally so we don't have to maintain copy-pasted code in several places.

PR #319, opened 2025-11-15
add alma10 images

h-vetinari:
Necessary for https://github.com/conda-forge/linux-sysroot-feedstock/pull/92, which has already been requested for a while, and will also be necessary for riscv64 support that was discussed in the last core call


xhochy (approved):

h-vetinari:
Thanks for the review!

